PR #13, opened 2022-11-10
fix: `insertMarkdownContent` didn't insert Markdown correctly in plain-text documents

rfgamaral:
## Overview

I've noticed that the `insertMarkdownContent` command was not working properly for plain-text documents, and this PR fixes that.

## PR Checklist

## Test plan

- Open the `Plain-text → Functions → Commands` story
- Click the `insertMarkdownContent` button
    - [x] Observe that Markdown content is inserted correctly (check the demo below)

## Demo

| image | image |


frankieyan (approved):
I can confirm this works in Storybook 👍

Just to double check, there is currently no use case for the `insertMarkdownContent` command with a plain text editor, correct?

rfgamaral:
> Just to double check, there is currently no use case for the `insertMarkdownContent` command with a plain text editor, correct?

Yeah, I don't think we are using it anywhere at the moment. Not sure if we ever will, but it was a quick bugfix 😅

PR #40, opened 2022-12-02
ci: Implement CD on every push/merge to `main`

rfgamaral:
## Overview

Updates the `Typist Package Release` workflow to trigger a new package release with every push/merge to `main`.

This takes advantage of the `workflow_run` event to only run after the `CI Validation` workflow on `main` is completed. If the `CI Validation` on `main` fails, the `Typist Package Release` workflow won't run. Similarly, if the `CI Validation` on `main` is cancelled (for instance, due to a subsequent PR merge), the `Typist Package Release` workflow will be skipped for that run due to the `github.event.workflow_run.conclusion == 'success'` condition check.

P.S: No need to worry about pushes/merges to `main` that don't change the package output (like documentation changes), because `semantic-release` is aware of that based on the Conventional Commits specification, and only `fix` or `feat` commits will publish a new package.

## PR Checklist

## Test plan

The only way to properly test this change is to merge the PR, and see if a new release is published (it should). Subsequent merged PRs should keep triggering new releases if the commit type is either `fix` (patch release), `feat` (minor release), or any commit type with a `!` appended after the type/scope (breaking change, major release).

### DIFF
--- a/.github/workflows/publish-typist-package-release.yml
+++ b/.github/workflows/publish-typist-package-release.yml
@@ -1,46 +1,30 @@
 name: Typist Package Release
 
 on:
-    workflow_dispatch:
+    workflow_run:
+        workflows:
+            - CI Validation
+        branches:
+            - main
+        types:
+            - completed
+
+# The release workflow involves many crucial steps that once triggered it shouldn't be cancelled
+# until it's finished, otherwise we might end up in an inconsistent state (e.g., a new release
+# published to npm but not GitHub Packages). To prevent this, concurrency is disabled with
+# `cancel-in-progress: false`, and new workflow runs will be queued to be started only when the
+# previous one has completely finished.
+concurrency:
+    group: typist-package-release
+    cancel-in-progress: false
 
 jobs:
-    prepare-workflow:
-        name: Prepare Workflow
-        runs-on: ubuntu-latest
-        timeout-minutes: 5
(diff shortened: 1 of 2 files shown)


.github/workflows/publish-typist-package-release.yml
```diff
@@ -82,38 +67,22 @@ jobs:
                   GIT_COMMITTER_NAME: Doist Bot
 
             - name: Remove Doist registry configuration from `.npmrc`
+              if: ${{ steps.semantic-release.outputs.package-published == 'true' }}
```
scottlovegrove, line 70: This is either a valid question, or a learning opportunity for me, but where are these being set to true? I can't see anything setting them to be true, so will they always be false?

rfgamaral: `semantic-release` is governed by the `.releaserc` file, as the last step, when a release is successful, we change `package-published` to `true` ([ref](https://github.com/Doist/typist/pull/40/files#diff-644a8fae29c8e8b6ad5405a2e66c718a3b27f2e12581b7c5e00396d32b3e5cdeR32)).

scottlovegrove: Ah nice, learning moment for me then :)

scottlovegrove (approved):

PR #66, opened 2022-12-15
fix(factories): Allow alphanumeric IDs for suggestion nodes in `createSuggestionExtension`

### LINKED ISSUE #64: How to modify the selected suggestion node
Right now, we receive the markdown output of an `@mention` node as `[mentioned_name](mention://ID)` (`ID` being a `number`).
For my use case, I need
1. An alphanumeric string as ID (Doing so currently results in `NaN` in the .md output )
2. ~~Replace the node with a custom node or provide a custom node in config~~
~~Can someone please let me know how to achieve the goals listed above, Thanks.~~

References:
Where ID is given to the item: https://github.com/Doist/typist/blob/416fb2655becbc7228891fd583ae6fd3aa470189/stories/typist-editor/extensions/suggestions/mention-suggestion-dropdown.tsx#L22

Where item suggestion item created: https://github.com/Doist/typist/blob/main/src/factories/create-suggestion-extension.ts

rfgamaral:
## Overview

Fixes #64.

## PR Checklist

-   [x] Changes introduced here have been manually tested by someone other than the PR author
-   [x] Added/updated documentation to Storybook or `README.md`

## Test plan

- Add `[John Doe](mention://@user123:matrix.org)` in the `content` control (bottom)
    - [x] Observe that the editor was initialized with a mention suggestion identified
    - [x] Observe that the Markdown output shows `[John Doe](mention://@user123:matrix.org)`


frankieyan (approved):
Works well 👍 

I suppose this would be a minor rather than breaking change as this doesn't change the markdown serialization, and we can continue to use numeric ids as they would be converted to strings internally by the extension, correct?

rfgamaral:
> I suppose this would be a minor rather than breaking change as this doesn't change the markdown serialization, and we can continue to use numeric ids as they would be converted to strings internally by the extension, correct?

Yes, when `storage.itemsById` is populated, keys are converted to strings.

---

But I'm glad you asked this because it prompted me to review the `attributesMapping` constraints in `createSuggestionExtension`, and I've realized that I broke the mapping for `number` properties in this PR. To explain what I mean, image `SuggestionItem` is defined like this:

```ts
type SuggestionItem = {
    uid: number
    name: string
}
```

But, internally, the extension uses `id`, so we need to map one property to the other, like this:

```ts
createSuggestionExtension<SuggestionItem>(MENTION_SUGGESTION_TYPE, MENTION_SUGGESTION_ITEMS, {
    id: 'uid',
    label: 'name',
})
```

However, with the changes I made in this PR, this doesn't work:

image

Because we are trying to map a `number` (i.e. `uid`), to a `string` (i.e. both `id` and `label` only allow strings).

So, technically, this is a breaking change, but it wouldn't actually break anything **for us** because our apps use `id` by default, and we don't need to map this attribute in particular, and since the extension will take care of converting everything to strings, everything just keeps working.

However, there's no reason to introduce this breaking change, even if it doesn't affect us. It's a simple fix, and doesn't break anything for anyone. I've also updated the `MentionSuggestion` and `HashtagSuggestion` extensions to serve as examples for different implementations.

Just thought I'd leave this whole clarification here for posterity.

rfgamaral:
@frankieyan Made another commit, shouldn't change anything, but please double-check 🙏

frankieyan:
Thanks for the explaination 👍 both hashtags and mentions are still working in storybook 🙌

frankieyan (approved):

PR #72, opened 2022-12-19
fix(rich-text-link): More lenient regex for input/paste rule

rfgamaral:
## Overview

This will help with Doist/Issues#8630.

This fixes the aforementioned issue by having a more lenient RegExp for the `RichTextLink` extension input/paste rules. Instead of having a strict RegExp, where we only allow certain characters, we now allow pretty much anything except for whitespace characters (i.e. `\S`). This way, it's up to the user to be sure they are using a valid URL, instead of placing the onus on the editor.

## PR Checklist

## Test plan

- Paste the following in the `content` control (at the bottom):
    ```
    [Link-1](doist.dev

    [Link-2](https://doist.dev

    [Link-3](https://doist.dev/example

    [Link-4](https://doist.dev/example/file.html

    [Link-5](https://doist.dev/example/file.html?foo

    [Link-6](https://doist.dev/example/file.html?foo=bar

    [Link-7](https://doist.dev/example/file.html?foo=bar&baz=qux

    [Link-8](https://doist.dev/example/foo-bar.html?baz=qux%20quux
    ```
    - Alternatively, paste directly into the editor with `Cmd/Ctrl+Shift+V`
- For each link in the editor, type `)` at the end
    - [ ] Observe that that Markdown output is exactly what you would expect

### DIFF
--- a/src/extensions/rich-text/rich-text-link.ts
+++ b/src/extensions/rich-text/rich-text-link.ts
@@ -6,18 +6,14 @@ import type { LinkOptions } from '@tiptap/extension-link'
 /**
  * The input regex for Markdown links with title support, and multiple quotation marks (required
  * in case the `Typography` extension is being included).
- *
- * @see https://stephenweiss.dev/regex-markdown-link
  */
-const inputRegex = /(?:^|\s)\[([^\]]*)?\]\(([A-Za-z0-9:/.-?]+)(?: ["“](.+)["”])?\)$/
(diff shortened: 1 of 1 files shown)


gnapse (approved):
Looks good. My only observation is: can't we have tests that would check the same as the test plan does?

rfgamaral:
> Looks good. My only observation is: can't we have tests that would check the same as the test plan does?

@gnapse We can, and we should! But these need to be end-2-end, and I didn't manage to find the time in my Personal DO to research the best way to implement those. It's still on my to-do list, and something I'll tackle within my housekeeping time. This is definitely not the state that I want us to be in, but for now this will have to do.

PR #86, opened 2023-01-05
fix(html-serializer): Disables tokenizers if marks/nodes are not found in the editor schema

rfgamaral:
This commit simply reviews some comments/JSDoc and rewrites the code
structure a bit (not introducing breaking changes) to allow for future
extensibility to the serializer object.

## Overview

This will help fix an internal issue where the `RichTextKit` is used with a bunch of extensions disabled, but since the HTML serializer does not take into account all marks/nodes that are unavailable (due to the disabled extensions), the Markdown input is still parsed as if most extensions were enabled. Because of this, the HTML output will contain markup that the editor cannot handle. thus not rendering all content as expected.

## PR Checklist

## Test plan

This PR is a bit hard and time-consuming to test because it requires changing Storybook code to replicate the behaviour that we have on Todoist, and it would take _a lot of time_ to write a thorough test plan. However, I took some time to review the serializers unit tests to make sure that everything was tested properly. Please focus your review in the code changes only (and take a special look at the commit descriptions to better understand the "refactor" changes).


craigcarlyle (approved):
@rfgamaral Code changes here look reasonable and no issues discovered during smoke testing 👍

PR #88, opened 2023-01-06
feat(serializers): Add `get*` functions for reusable singular instances

rfgamaral:
## Overview

For a while now that I've been bothered by the default that we were recreating the Markdown/HTML serializer every time that we needed them (for the `<TypistEditor>` component, built-in extensions, and custom extensions built exclusively for our products). With this PR I'm introducing `getMarkdownSerializerInstance` and `getHTMLSerializerInstance`, which work in tandem with `createMarkdownSerializer` and `createHTMLSerializer`, respectively.

The goal of these new `get*` functions is to return a singleton instance for each of the serializers based on provided editor schema argument, preventing the serializer from being recreated every time that we need to use it. Technically, this is a more efficient solution than what we had before, but in practice we are not likely to notice any performance improvements.

## PR Checklist

-   [x] Added/updated documentation to Storybook or `README.md`

## Test plan

This PR doesn't add any new functionality nor fixes anything per se, so a thorough code review should suffice. Once this is merged and a new Typist release is published, I will update `todoist-web` and `twist-web`, and make sure that both are using the new `get*` functions.

> **Note**
> The implementation of these singleton `get*` functions have been extensively discussed and tested between me and Pedro, and this is the final implementation that we both agreed on. However, in the spirit of knowledge sharing, I'm making this a `ask` PR so that someone else on the team becomes more familiar with Typist.


src/serializers/html/html.ts
```diff
@@ -135,6 +142,28 @@ function createHTMLSerializer(schema: Schema): HTMLSerializerReturnType {
...
+/**
+ * [[DESCRIPTION]]
+ *
+ * @param id [[DESCRIPTION]]
+ * @param schema [[DESCRIPTION]]
+ *
+ * @returns [[DESCRIPTION]]
+ */
+// TODO: Needs unit tests!
+function getHTMLSerializerInstance(id: NonEmptyString, schema: Schema) {
```
rfgamaral, line 159: 💭 Given that the `TypistEditor.id` prop is set to `default` in case the user doesn't specify one, it could make sense to switch the parameters and make `id` optional (defaulting to `default` too).

nats12 (approved):
Apologies for the delay, LGTM. Thanks for sharing the PR 👍🏼 kudos for the documentation around the new functions.

PR #98, opened 2023-01-13
fix: Replace usage of useEvent with useCallback

pawelgrimm:
## Overview

A function created with the `useEvent` hook was being used in `TypistEditor`'s `onCreate` function, which was called during rendering (when initializing the `editor` `useState` in `useEditor`). This caused it to throw an error when React's Strict Mode was enabled, as this type of usage makes the rendering result non-determinisitic. See https://github.com/reactjs/rfcs/pull/220#issuecomment-1118055107 for further discussion.

## PR Checklist

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -103,6 +103,7 @@
                 "react-icons": "4.7.1",
                 "react-markdown": "8.0.5",
                 "react-syntax-highlighter": "15.5.0",
+                "react-use-event-hook": "0.9.3",
                 "rehype-raw": "6.1.1",
                 "remark-gfm": "3.0.1",
                 "rimraf": "3.0.2",
@@ -125,7 +126,6 @@
                 "marked": "^4.1.1",
                 "react": "^17.0.0 || ^18.0.0",
                 "react-dom": "^17.0.0 || ^18.0.0",
-                "react-use-event-hook": "^0.9.3",
                 "turndown": "^7.1.1"
             }
         },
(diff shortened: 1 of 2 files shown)


rfgamaral:
> Can you share any context around why we opted for `useEvent` here in the first place?

Before using `useEvent` we were using `useRefCallback` ([ref](https://github.com/Doist/frontend-toolbox/blob/main/packages/react-hooks/src/hooks/useRefCallback.ts)), which was our own custom version of `useEvent`. We moved to `useEvent` when Typist became open-source, and to lose the dependency on the private `@doist/toolbox-react-hooks` package.

With that said, I was first introduced to `useRefCallback` by Ernesto in [this thread](https://twist.com/a/1585/ch/190200/t/3006188/), and although that was mainly a question for event handlers **passed on to** `<TypistEditor>` (at the time it was called `RichTextEditor`), I somehow felt the need to use it in the `onCreate` handler **inside** `<TypistEditor>`.

I don't have more context than that, and `useCallback` might be sufficient here, however, I have some questions before we consider merging this, but I believe https://github.com/Doist/todoist-web/pull/5048 to be the more appropriate place to discuss that.

rfgamaral:
@pawelgrimm I honestly cannot say if that will be the case or not. I don't think we'll introduce weird behaviour with this change, because like I said, I added it here "just because", no good reason, and probably shouldn't have.

However, to avoid going back and forth with Typist releases (and updating Todoist), may I suggest that we apply a patch on `todoist-web` with this change and validate that way? We can have one (or more) people from the team validate the PR, and then we'll also have our users validating the change for us. If we don't get any tickets in a day, we are probably safe 😅

If you go that route, please be careful with the current patch, and keep it around, it's still needed.

rfgamaral (dismissed):
Since `useEvent` is no longer used on Typist itself, we also need to move `"react-use-event-hook": "^0.9.3",` from `peerDependencies` to `devDependencies` (don't forget to remove the `^` from the version number when moving it to `devDependencies`, and make sure the alphabetical order is kept).

rfgamaral:
@pawelgrimm I think we can proceed with this, don't you think?

As a reminder, don't forget to do the change I requested above, and also change the PR title from `refactor:` to `fix:`. Technically, it's a fix, right? Either way, `refactor:` won't trigger a new release automatically, and `fix:` will, which is what we want after we merge this.

Then feel free to update Todoist with the new version (and don't forget to update the revert the patch to the previous version).

rfgamaral (approved):

PR #102, opened 2023-01-16
fix(markdown-serializer): Override Turndown escaping behaviour with custom rules

rfgamaral:
## Overview

We've detected a few issues and inconsistencies with how our Markdown serializer escapes Markdown syntax, and this PR aims to fix that. The escaping is done by a bunch of aggressive built-in rules (based on regular expressions) implemented in Turndown, and this PR overrides the `escape` function with a custom implementation with custom rules.

We've been discussing this internally ([ref](https://twist.com/a/1585/ch/190200/t/4031706/)), and this is the best solution I could come up with. However, I'm going to keep this in a draft state until we do a live test directly in production 🔥 (no better way to test this, really).

## PR Checklist

## Test plan

Since this will be tested directly in production, and given that unit tests have been written to make sure the Markdown serializer outputs exactly what's expected, a code-review should suffice.


rfgamaral:
A patch of this PR has been in production for at least 3 weeks now, and no major bugs have been reported. I believe this fix to be in an optimal state to be merged as-is, and will do just that once all statuses checks are green.

PR #163, opened 2023-03-15
fix(markdown-serializer): Add custom rule to escape horizontal rules for rich-text

rfgamaral:
## Overview

With #102 we overrode Turndown's escape mechanism with our custom implementation with custom rules to overcome the agressive rules built-in the default implementation. Before making this a permanent change, we tested the custom implementation in production for more than a month, and since no bugs were reported during that time, we decided to keep it. However, caused by this custom escape mechanism, we recently got our first bug report ([ref](https://github.com/Doist/Issues/issues/9658)). This PR fixes that issue by adding a custom rule to the custom escape mechanism.

## PR Checklist

## Test plan

The new custom rule is already fully tested by the added unit test, and unfortunately it's not possible to manually test this without making changes to the Storybook editor, which makes for a cumbersome test plan.

### DIFF
--- a/src/serializers/markdown/markdown.test.ts
+++ b/src/serializers/markdown/markdown.test.ts
@@ -636,6 +636,45 @@ See the section on [\`code\`](#code).`,
                     ).toBe(`- 1968\\. A great year!
 - I think 1969 was second best.`)
                 })
+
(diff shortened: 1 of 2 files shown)


rfgamaral:
A simpler and less hacky solution was implemented on the renderer side instead ([ref](https://github.com/Doist/todoist-web/pull/6214)).

PR #260, opened 2023-05-12
fix(paste-markdown): Incorrect paste behaviour when HTML source is VSCode

rfgamaral:
## Overview

This fixes a long-standing issue when pasting rich-text (HTML) content copied from VS Code while using a Chromium based browser (other browsers don't have support to detect VS Code as the source of the content like Chromium does).

> **Note**
> I took the opportunity to refactor the `PasteMarkdown` extension slightly to make the code more clear and readable, but the downside is that it makes it harder to review the actual changes that fix the issue. With that in mind, I left comments in PR explaining what were the actual changes, everything else is pure refactor.

## PR Checklist

### Preparation

- Open VS Code, create a new Markdown file
    - Make sure that `Language Mode` is set to `Markdown`
- Copy the following content and paste it into the Markdown file in VS Code:
    ```
    # Qui cum ad ignara intremuit volucres formidine

    ## Quidem pectus

    Lorem markdownum terraeque deos nocte novissima ego nepotem fraternis gaudere;
    ubi. Quem sim de sparsi auctor, est et an Iovis me lilia germanamque.

    - Lacrimaeque suaque ab inquit quos stolidarum omnia
    - Peregrina est nubes in levavit

    Sicaniam merguntque mentior vicem magis latices, et amara et manu expulit,
    venit. Et iube tamen verba et nihil [amantis dum
    vitibus](http://stat.com/etet.html) Ditis ad!
    ```

### Core Test #1

- Open the [preview Storybook](https://deploy-preview-260--doist-typist.netlify.app/?path=/docs/readme--docs) deployed to Netlify **on a Chromium based browser**
- Copy the Markdown content that you pasted into VS Code from VS Code
    - **It's important that the copied content comes from VS Code (and not GitHub)**
- Paste it into the Rich-text editor on the preview Storybook with `Cmd/Ctrl+V`
    - [x] Observe that the Markdown content is pasted with correct formatting
        - If you compare this with [production](https://typist.doist.dev/?path=/story/typist-editor-rich-text--default), you'll see that formatting is lost
- Now, do the same but use the `Cmd/Ctrl+Shift+V` shortcut instead
    - [x] Observe that the Markdown content is pasted with correct formatting (exactly as before)

### Bonus Test

If you want to try this on Firefox/Safari, here's what you should expect...

- Open the [preview Storybook](https://deploy-preview-260--doist-typist.netlify.app/?path=/docs/readme--docs) deployed to Netlify **on a Chromium based browser**
- Copy the Markdown content that you pasted into VS Code from VS Code
    - **It's important that the copied content comes from VS Code (and not GitHub)**
- Paste it into the Rich-text editor on the preview Storybook with `Cmd/Ctrl+V`
    - [x] Observe that the Markdown content is NOT pasted with correct formatting
        - If you compare this with [production](https://typist.doist.dev/?path=/story/typist-editor-rich-text--default), you'll get the same result
        - As explained before, this is expected, because other browsers don't have support to detect VS Code as the source of the content like Chromium does
- Now, do the same but use the `Cmd/Ctrl+Shift+V` shortcut instead
    - [x] Observe that the Markdown content is pasted with correct formatting
        - In this case it's expected to work because with this shortcut we are forcing the plain-text version from the clipboard data to be pasted instead, and we always handle this ourselves

### DIFF
--- a/src/extensions/rich-text/paste-markdown.ts
+++ b/src/extensions/rich-text/paste-markdown.ts
@@ -37,9 +37,9 @@ const PasteMarkdown = Extension.create({
                         return Slice.maxOpen(Fragment.from(editor.schema.text(text)))
                     },
                     handlePaste(_, event, slice) {
-                        const clipboardText = event.clipboardData
-                            ?.getData(ClipboardDataType.Text)
-                            .trim()
+                        // The clipboard contains text if the slice content size is greater than
+                        // zero, otherwise it contains other data types (like files or images)
+                        const clipboardContainsText = Boolean(slice.content.size)
 
                         const clipboardContainsHTML = Boolean(
                             event.clipboardData?.types.some(
@@ -49,7 +49,7 @@ const PasteMarkdown = Extension.create({
 
                         // Unfortunately, the VS Code clipboard data type is not supported by
                         // Firefox or Safari, which means that copy/paste experience from VS Code
-                        // into the editor with any of those browsers is supbar:
+                        // into the editor with either of those browsers is subpar:
                         //  * The Markdown syntax is not fully converted to rich-text
                         //  * Code is not detected nor converted to a code-block
                         const clipboardContainsVSCodeMetadata = Boolean(
@@ -57,6 +57,9 @@ const PasteMarkdown = Extension.create({
                                 (type) => type === ClipboardDataType.VSCode,
                             ),
                         )
+
+                        const clipboardContainsHTMLFromUnknownSource =
+                            clipboardContainsHTML && !clipboardContainsVSCodeMetadata
 
                         const vsCodeClipboardMetadata: VSCodeClipboardMetadata =
                             clipboardContainsVSCodeMetadata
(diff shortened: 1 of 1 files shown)


src/extensions/rich-text/paste-markdown.ts
```diff
@@ -66,27 +69,42 @@ const PasteMarkdown = Extension.create({
                                   ) as VSCodeClipboardMetadata)
                                 : {}
 
+                        const clipboardContainsHTMLFromVSCodeOtherThanTextOrMarkdown =
+                            clipboardContainsVSCodeMetadata &&
+                            // If `mode` from the VS Code metadata is `null` it probably means that
+                            // the user has the VS Code `editor.copyWithSyntaxHighlighting` setting
+                            // set to `false`, thus returning plain-text
+                            vsCodeClipboardMetadata.mode !== null &&
```
rfgamaral, line 77: This is one of the functional changes, which makes pasting content copied from VS Code work in Chromium browsers regardless if users have `editor.copyWithSyntaxHighlighting` set to `true` (default if not set), or `false`.

src/extensions/rich-text/paste-markdown.ts
```diff
@@ -66,27 +69,42 @@ const PasteMarkdown = Extension.create({
...
                         // Send the clipboard text through the HTML serializer to convert potential
                         // Markdown into HTML, and then insert it into the editor
                         editor.commands.insertMarkdownContent(
-                            // The slice content is used instead of `clipboardText` as the pasted
-                            // content could have already been transformed by other plugins
-                            slice.content.textBetween(0, slice.content.size),
+                            // The slice content is used instead of getting the text directly from
+                            // the clipboard data because the pasted content could have already
+                            // been transformed by other ProseMirror plugins
+                            slice.content.textBetween(0, slice.content.size, '\n'),
```
rfgamaral, line 107: This was the other functional change, the main one actually, which makes sure that multi-line Markdown copied from VS Code work is split correctly (in Chromium browsers) by inserting aline feed character between nodes.

Bloomca (approved):
Works as expected in Chrome 👍 

It also works as expected in Firefox on Windows for me, even though it seems it shouldn't. But it works this way before the change as well, so no changes there.

rfgamaral:
That's strange, @Bloomca.

Here's a comparison betweeb Firefox and Edge (Chormium) on my side running the deployed preview above:

https://github.com/Doist/typist/assets/96476/2c0eb598-40bc-4452-9c98-8e4dcc6cb5f0

Is this different from what you get?

rfgamaral:
That's strange, @Bloomca.

Here's a comparison betweeb Firefox and Edge (Chormium) on my side running the deployed preview above:

https://github.com/Doist/typist/assets/96476/759c34f2-08c9-4edc-8e97-50eb438f1383


Is this different from what you get?

Bloomca:
@rfgamaral sorry, I missed that one.

In Firefox, both on this branch and the production one, when I paste using `Ctrl + V`, I see this:

image

`Ctrl + Shift + V` works the same way.

In Chromium, it works incorrectly on production, but correctly on this branch.

Bloomca:
Ricardo debugged it with me in DMs, and turned out that in my VSCode I had this setting set to `false`:

```
"editor.copyWithSyntaxHighlighting": false,
```

After I switched to `true` (the default value), it works as expected in Firefox now.

PR #275, opened 2023-05-29
fix(html-serializer): Don't share instances between editors

rfgamaral:
## Overview

This PR fixes an internal issue on Todoist which is caused by the fact that `marked` shares a single instance (with the same options and extensions) across all usages. This is problematic for us because we have multiple editors in a single page with different options/extensions. That issue is fixed by replacing the HTML serializer based on `marked` with the [`unified` ecosystem](https://unifiedjs.com/) instead.

## PR Checklist

## Test plan

I did my best to carefully test the changes by reviewing all unit tests and making changes where appropriate, and I also pulled in these changes into Todoist to make sure that the issued that is supposed to fix, it's actually fixed (spoiler: it is 🥳). But the real test will be to push this into production, and see if it breaks anything. Hopefully it won't 🤞.

You can, however, play with the [rich-text editor preview below](https://deploy-preview-275--doist-typist.netlify.app/?path=/story/typist-editor-rich-text--default) by changing the `content` control, which takes Markdown as input, and goes through the new HTML serializer to output HTML ready for Tiptap.


nats12 (approved):
Looks good Ricardo 👏🏼  I played around with the deployment preview and saw no issues 👍🏼

PR #290, opened 2023-06-06
feat: Add the `PasteHTMLTableAsString` extension

rfgamaral:
## Overview

This PR introduces the `PasteHTMLTableAsString` extension, which adds the ability to paste a table copied from a spreadsheet web app (e.g., Google Sheets, Microsoft Excel), along with tables rendered by GitHub Flavored Markdown (GFM), into the editor.

This extension will help fix an internal issue where some users copy a list of rows from a spreadsheet web app and paste them into Todoist expecting to be prompted to add each row as a separate task (as if they were pasting multiple lines of text).

## PR Checklist

## Test plan

- Create a Google Sheet with a sample table (multiple columns/rows)
    - Copy and paste the table (or parts of it) into the editor
        - [x] Observe that each row is pasted into the editor as paragraph
        - [x] Observe that each column is separated by a ` ` character
- Create a Microsoft Excel document with a sample table (multiple columns/rows)
    - Copy and paste the table (or parts of it) into the editor
        - [ ] Observe that each row is pasted into the editor as paragraph
        - [ ] Observe that each column is separated by a ` ` character
- Create a Numbers (Apple) document with a sample table (multiple columns/rows)
    - Copy and paste the table (or parts of it) into the editor
        - [ ] Observe that each row is pasted into the editor as paragraph
        - [ ] Observe that each column is separated by a ` ` character

### GFM test

- Take the following Markdown table as input:
    | H1 | H2 | H3 |
    |--------|--------|--------|
    | _A1_ | A2 | A3 |
    | B1 | **B2** | B3 |
    | C1 | C2 | ~~C3~~ | 

    - Copy and paste the table (or parts of it) into the editor
        - [x] Observe that each row is pasted into the editor as paragraph
        - [x] Observe that each column is separated by a ` ` character
        - [x] Observe that formatting for the formatted cells is preserved

### Bonus tests

- You can try the same as above for the `Plain-text → Default` story, and most of the behaviour should be very similar. The small difference is that a tabular space (i.e. `\t`) will be used, instead of ` `) to separate columns.
- Regardless of the number of columns copied, if all cells within a row are empty, that row should be discarded (i.e. not even an empty paragraph will be shown)
- Feel free to think of other edge cases that some users might use, and let me know if something doesn't work as you'd expect

### Typist

https://github.com/Doist/typist/assets/96476/eb90d340-b16c-47d3-8579-483f5efe4cf4

### Todoist

https://github.com/Doist/typist/assets/96476/8cab1c4b-21b9-4c1d-9c7e-0a62db8981af

### DIFF
--- a/src/extensions/shared/paste-spreadsheet-table.ts
+++ b/src/extensions/shared/paste-spreadsheet-table.ts
@@ -1,0 +1,74 @@
+import { Extension } from '@tiptap/core'
+import { Plugin, PluginKey } from 'prosemirror-state'
+
+import { PASTE_EXTENSION_PRIORITY } from '../../constants/extension-priorities'
+import { parseHtmlToElement } from '../../helpers/dom'
+
+/**
+ * The `PasteSpreadsheetTable` extension adds the ability to paste a table copied from a spreadsheet
+ * web app (e.g., Google Sheets, Microsoft Excel), along with tables rendered by GitHub Flavored
+ * Markdown (GFM), into the editor.
+ *
+ * Since Typist does not yet support tables, this extension simply pastes the table as a string of
+ * paragraphs (one paragraph per row), with each cell separated by a space character. However,
+ * whenever we do add support for tables, this extension will need to be completely rewritten.
+ *
+ * Lastly, please note that formatting is lost when the copied table comes from Google Sheets or
+ * Microsoft Excel, because unfortunately, these apps style the cell contents using CSS.
+ */
+const PasteSpreadsheetTable = Extension.create({
+    name: 'pasteSpreadsheetTable',
+    priority: PASTE_EXTENSION_PRIORITY,
+    addProseMirrorPlugins() {
+        return [
+            new Plugin({
+                key: new PluginKey('pasteSpreadsheetTable'),
+                props: {
+                    transformPastedHTML(html) {
+                        // Attempt to extract a table HTML from the pasted HTML
+                        const tableHTML = /<table[^>]+>[\s\S]*?<\/table>/gi.exec(html)
+
+                        // Do not handle the event if a table was not found
+                        if (!tableHTML) {
+                            return html
+                        }
+
+                        const { firstElementChild: tableElement } = parseHtmlToElement(
+                            tableHTML[0],
+                        ) as {
+                            firstElementChild: HTMLTableElement | null
+                        }
+
+                        // Do not handle the event if we don't have a table element
+                        if (!tableElement) {
+                            return html
+                        }
+
+                        // Transform the table element into a string of paragraphs
+                        return (
+                            Array.from(tableElement.rows)
+                                // Join each cell into a single string for each row
+                                .reduce<string[]>((acc, row) => {
+                                    return [
+                                        ...acc,
+                                        Array.from(row.cells)
+                                            // Use `innerHTML` instead of `innerText` to preserve
+                                            // potential formatting (e.g., GFM) within each cell
+                                            .map((cell) => cell.innerHTML)
+                                            .join(' '),
+                                    ]
+                                }, [])
+                                // Discard rows that are completely empty
+                                .filter((row) => row.trim().length > 0)
+                                // Wrap each row in a paragraph
+                                .map((row) => `<p>${row}</p>`)
+                                .join('')
+                        )
+                    },
+                },
+            }),
+        ]
+    },
+})
+
+export { PasteSpreadsheetTable }
--- a/src/constants/extension-priorities.ts
+++ b/src/constants/extension-priorities.ts
@@ -13,6 +13,13 @@ const SUGGESTION_EXTENSION_PRIORITY = 1000
 const SMART_MARKDOWN_TYPING_PRIORITY = 110
 
 /**
+ * Priority for the `PasteSpreadsheetTable` extension. This needs to be higher than most paste
+ * extensions (e.g., `PasteSinglelineText`, `PasteMarkdown`, etc.), so that the extension can first
+ * parse HTML tables that might exist in the clipboard data.
+ */
+const PASTE_EXTENSION_PRIORITY = 105
+
+/**
  * Priority for the `ViewEventHandlers` extension. This needs to be higher than the default for most
  * of the built-in and official extensions (i.e. `100`), so that the event handlers from the
  * extension can take precedence over the built-in and official extensions event handlers.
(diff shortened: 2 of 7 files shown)


src/extensions/shared/paste-spreadsheet-table.ts
```diff
@@ -0,0 +1,74 @@
...
+                        // Attempt to extract a table HTML from the pasted HTML
+                        const tableHTML = /<table[^>]+>[\s\S]*?<\/table>/gi.exec(html)
+
+                        // Do not handle the event if a table was not found
+                        if (!tableHTML) {
+                            return html
+                        }
+
+                        const { firstElementChild: tableElement } = parseHtmlToElement(
+                            tableHTML[0],
```
frankieyan, line 37: I tried copying two tables (from GitHub) and it only processed the first table. For example, this selection would only result in `C1 C2 ~~C3~~`:

image

Before this change, the output would be:

```
C1C2~~C3~~

H2H2H3_A1_A2A3
```

Should we support the case of multiple tables as well?

src/extensions/shared/paste-spreadsheet-table.ts
```diff
@@ -0,0 +1,74 @@
...
+
+import { PASTE_EXTENSION_PRIORITY } from '../../constants/extension-priorities'
+import { parseHtmlToElement } from '../../helpers/dom'
+
+/**
+ * The `PasteSpreadsheetTable` extension adds the ability to paste a table copied from a spreadsheet
+ * web app (e.g., Google Sheets, Microsoft Excel), along with tables rendered by GitHub Flavored
+ * Markdown (GFM), into the editor.
+ *
+ * Since Typist does not yet support tables, this extension simply pastes the table as a string of
```
frankieyan, line 12: Do you think it would be helpful to be more explicit in the extension's name (e.g. pasteSpreadsheetTableAsParagraphs)? It might be easier to figure out at a glance what's supported, but if we ever do support markdown tables in the future, then we could reuse this name while re-implementing its internals.

frankieyan (approved):
Looks good, @rfgamaral. Other than with multiple windows (which isn't possible in a spreadsheet application, so just markdown, I think), I can't find anything wrong.

I don't have excel, so I tested Numbers instead, and for some reason it favours pasting an image over html. Here is the clipboard's data: [image](https://github.com/Doist/typist/assets/8531248/c7a2acdd-dbf8-45a0-8a23-3500bc44e9e1)

src/extensions/shared/paste-spreadsheet-table.ts
rfgamaral: It wasn't too hard or time consuming, so I added support for multiple (GFM) tables.

src/extensions/shared/paste-spreadsheet-table.ts
rfgamaral: I renamed the extension to be more explicit on what it does.

However, it got me thinking about this comment that I wrote, which doesn't make much sense... If we ever add proper table suppor to the editor, this extension **should not** be rewritten as I first through, because we are still gonna need the functionality that it provides for editors where tables aren't supported (i.e. Todoist task title). Instead, what we need to do is either enable/disable this extension when required.

With that in mind, I'm thinking if it wouldn't make sense to revert the way it's included in the rich-text/plain-text kits. I mean, right now I added them by default to both kits, and we can disable them per editor. Would it make sense to have this extension not added by default to both kits, and with the option enable them per editor?

Also, we don't necessarily need to have it enable/disable by default on both the rich-text/plain-text (one can be enabled by defualt while we disable by default on the other), because for rich-text we might add table support down the road and have it turned on by default, but for plain-text we'll never do that, because, well, it's plain-text (at most we'd need a new extension that would transform an HTML table into a Markdown one, but that's a different feature altogether).

What do you think? Does it make sense that such extension is enabled by default or considered an extra that we enable when we need it? Should we have the same behaviour for both kits, or different?

rfgamaral:
@frankieyan Renamed the extension, and added support for multiple (GFM) tables, please take a second look when you can.

---

> Looks good, @rfgamaral. Other than with multiple windows (which isn't possible in a spreadsheet application, so just markdown, I think), I can't find anything wrong.

Not sure I understand what you mean here...

> I don't have excel, so I tested Numbers instead, and for some reason it favours pasting an image over html. Here is the clipboard's data: [image](https://github.com/Doist/typist/assets/8531248/c7a2acdd-dbf8-45a0-8a23-3500bc44e9e1)

Hum, interesting one... I have so many questions and things I'd like to understand, maybe we can take a few minutes from our 1:1 tomorrow to dig a bit into this since I don't have access to Numbers?

src/extensions/shared/paste-spreadsheet-table.ts
frankieyan: IMO it makes sense to enable this by default. I can't think of a use case where users would want to paste in spreadsheets cells that they've copied and have them all mashed together. We only use the single-line plain text kit in Twist, and I think in that case it's still better for the spaces to appear between cells.

When/if we do support markdown tables, I think this would still be a nice fallback when it's disabled. So my vote would be to have this enabled by default now for both kits 👍

frankieyan:
> > Looks good, @rfgamaral. Other than with multiple windows (which isn't possible in a spreadsheet application, so just markdown, I think), I can't find anything wrong.
> 
> Not sure I understand what you mean here...

Lol, "tables" was what I meant 🤦 
 
> Hum, interesting one... I have so many questions and things I'd like to understand, maybe we can take a few minutes from our 1:1 tomorrow to dig a bit into this since I don't have access to Numbers?

For sure. I do see `event.clipboardData.files` in the clipboard data from Numbers, which isn't present when you copy from Google Sheets. If we decide that extracting the text from the table is more important than the image, perhaps we need a `handlePaste` handler in this plugin and return whether a table was detected?

rfgamaral:
@frankieyan I fixed the issue where Numbers pasted an image instead of the text version of the table with the solution we discussed yesterday. Feel free to take a second look and test it on your side since I don't have a macOS device at my disposal.

However, I could not fix the Numbers table parsing, the HTML it outputs is really cumbersome to work with, and it would take a bit more effort to make it work (without breaking the current behaviour). We can revisit this in the future if it becomes a more prominent issue for our users.

frankieyan (approved):
Pasting from Numbers no longer pastes an image. The formatting does look way better/more accurate in the plain text kit vs the rich text kit, but let's go with this for now and see if there's any feedback from users.

PR #300, opened 2023-06-12
ci: Fix publishing to GitHub Packages registry

rfgamaral:
## Overview

It looks like https://github.com/Doist/typist/pull/284 wasn't enough to fix the `Typist Package Release` workflow, and publishing to GitHub packages still failed on the [latest release](https://github.com/Doist/typist/actions/runs/5241511888/jobs/9463707293).

After digging a bit deeper into this issue, I couldn't get provenance working with the GitHub Packages registry. It looks to me that provenance is not supported with the GitHub Packages registry, and my research about this led me nowhere. The solution I found was to enable provenance by default on `.npmrc` (to publish to npmjs with provenance), and then disable it before publishing to the GitHub Packages registry.

@scottlovegrove assigning this `show` PR to you for awareness, since you're the one involved in impelementing provenance in the first place.

## PR Checklist

## Test plan

As usual, this is the kind of change that can only be tested once the next release goes out. However, I personally used a forked repo to try the changes before pushing them into the real project, and I believe these changes to be okay.

### DIFF
--- a/.github/workflows/publish-typist-package-release.yml
+++ b/.github/workflows/publish-typist-package-release.yml
@@ -13,8 +13,11 @@ permissions:
     # Enable the use of OIDC for npm provenance
     id-token: write
     # Enable the use of GitHub Packages registry
-    contents: read
     packages: write
+    # Enable `semantic-release` to publish a GitHub release and post comments on issues/PRs
+    contents: write
+    issues: write
+    pull-requests: write
 
 # The release workflow involves many crucial steps that once triggered it shouldn't be cancelled
 # until it's finished, otherwise we might end up in an inconsistent state (e.g., a new release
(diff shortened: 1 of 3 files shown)


.github/workflows/publish-typist-package-release.yml
```diff
@@ -13,8 +13,11 @@ permissions:
     # Enable the use of OIDC for npm provenance
     id-token: write
     # Enable the use of GitHub Packages registry
-    contents: read
     packages: write
+    # Enable `semantic-release` to publish a GitHub release and post comments on issues/PRs
+    contents: write
+    issues: write
+    pull-requests: write
```
rfgamaral, line 20: These changes are unrelated to this PR, and `semantic-release` was fully working before without these permissions, however, the documentation states that these are the required permissions for every feature to work, so I'm adding them for good measure.

scottlovegrove (approved):

Issue #305, opened 2023-06-14
Links are lost on paste when their linked text include inline code


frankieyan:
Take this markdown for example:

```
[`Hello` there](https://www.google.ca)
```

When it's being pasted into the rich text editor, the link itself is gone, leaving only the content behind when converted back to markdown:

```
`Hello` there
```

You could recreate the link using the link toolbar, but it results in something like this instead:

```
`Hello` [there](https://www.google.ca)
```

On GitHub, the inline code element is respected when making the whole string a link:

[`Hello` there](https://www.google.ca)

---

This can be seen in our weekly bug rotation thread:

![image](https://github.com/Doist/typist/assets/8531248/a11fb334-1c5b-4609-b434-936b3dc8bb29)

frankieyan:
Removing this from the frontend project, as we have an internal issue tracking this [here](https://github.com/Doist/Issues/issues/7767). If it gets picked up by the housekeeping rotation we can start there, however, I'll keep this open to track this from Typist.

While this isn't something supported by Tiptap out of the box, I got from Ricardo that it's something we can support in Typist itself.

PR #309, opened 2023-06-16
feat: Allow all marks to coexist with the Code mark

frankieyan:
## Overview

This allows other marks to co-exist with the inline `code` mark, allowing it to be rendered with formatting and links, like this: [**`click me`**](https://todoist.com)

## PR Checklist

Closes https://github.com/Doist/typist/issues/305

-   [ ] Added/updated unit test cases and/or end-to-end test cases
-   [ ] Added/updated documentation to Storybook or `README.md`

## Test plan

- Open the deploy preview link below, and go to the rich text kit's demo
- Paste this content into the editor using cmd + shift + v (paste markdown):
```
[Testing `some code` in a link](https://google.ca/)
```
- [ ] The entire line is wrapped in the link, and the inline block is rendered properly
- [ ] Type some text as an inline code, then use the formatting toolbar to apply bold, italics, and a link to it. Each option is applied correctly
- [ ] Select your linked inline code and its surrounding content, then hit cmd + shift + c. Paste what you've copied somewhere and ensure the resulting markdown is correct

## Demo

| image | image |


src/extensions/rich-text/code.ts
```diff
@@ -0,0 +1,10 @@
+import { Code } from '@tiptap/extension-code'
+
+const RichTextCode = Code.extend({
+    // Instead of not allowing other marks to coexist, allow everything so that
+    // inline code can be bolded, linked, etc
+    // See https://tiptap.dev/api/schema#excludes
+    excludes: Code.name,
```
rfgamaral, line 7: I believe the correct way to allow everything is with:

```suggestion
    excludes: '',
```

We don't really need to exclude `Code` from itself, no other mark does this. For instance, you don't see the `excludes` property in the [Bold mark](https://github.com/ueberdosis/tiptap/blob/main/packages/extension-bold/src/bold.ts). This is probaly handled internally, and there's no need to exclude the own mark.

src/extensions/rich-text/code.ts
rfgamaral, line 1: We should probably name the file `rich-text-code.ts` for a) to match the extension name and b) consistency with other similar extensions.

src/extensions/rich-text/code.ts
```diff
@@ -0,0 +1,10 @@
+import { Code } from '@tiptap/extension-code'
+
+const RichTextCode = Code.extend({
+    // Instead of not allowing other marks to coexist, allow everything so that
+    // inline code can be bolded, linked, etc
+    // See https://tiptap.dev/api/schema#excludes
```
rfgamaral, line 6: My initial thought was to have `excludes: 'bold italic strike'`, basically only allowing the link mark (which was what the initial issue, and our internal issue, was about). However, this will provide us with a behavior more close to GitHub's, which is a good thing (from my point of view 😅).

Anyway, for documentation and consistency purposes, I propose the following JSDoc:

```suggestion
/**
 * Custom extension that extends the built-in `Code` extension to allow all marks (e.g., Bold,
 * Italic, and Strikethrough) to coexist with the `Code` mark (as opposed to disallowing all any
 * other mark by default).
 *
 * @see https://tiptap.dev/api/schema#excludes
 */
const RichTextCode = Code.extend({
```

rfgamaral: Also, it's probably worth checking if our Markdown styles (I believe they now live on Reactist, under the `<Prose>` component) have proper support to style inline code marks with surrounding bold, italic and strikethrough marks.

rfgamaral (commented):
Can't wait for this to be merged and the change to be applied to Twist, good job 👍

Besides all the comments I left in the code, I'd like to propose a slightly different PR title (which is what will be added to `CHANGELOG.md`):

- `feat: Allow all marks to coexist with the Code mark`

rfgamaral (changes requested):
I just realized that allowing all marks with the `Code` mark creates a small issue when applying marks programatically (through keyboard shortcuts or a toolbar). See how `Code` wraps around some marks like `Italic` and `Strike`, when it shouldn't:

firefox_l8oVwekirU

Although the style in the editor looks acceptable, inspecting the DOM reveals that the `<em>` and `<s>` elements are inside `<code>`, thus the Markdown output that you see above.

And as you can see from the demo, the order the toolbar buttons are pressed is not relevant, because the order they are applied come from the `priority` and/or the alphabetical mark name (when they have the same priority).

To fix this, I propose the following to be added to `extension-priorities.ts`:

```ts
/**
 * Priority for the `RichTextCode` extension. This needs to be lower than the default for most
 * built-in and official extensions (i.e. `100`), so that other marks wrap the `Code` mark, and not
 * the other way around (i.e. prevents `<code><em>code</em></code>` from happening).
 */
const CODE_EXTENSION_PRIORITY = 99
```

And `priority: CODE_EXTENSION_PRIORITY` be added to `RichTextCode` (before `excludes`, for consistency 😅).

src/extensions/rich-text/code.ts
frankieyan: I was reading prose mirror's [documentation](https://prosemirror.net/docs/ref/#model.MarkSpec.excludes) which mentions that it defaults to marks of the same type, but I guess this behaviour is built in and not affected by the value here 👍

src/extensions/rich-text/code.ts
frankieyan: We may also need to apply the highlight colour correctly to the hover styles, but the idle state looks correct:

image

image

frankieyan:
Thank you Ricardo, nice find with mark ordering bug 🙌

Issue #423, opened 2023-08-30
Pasting a URL on a text selection no longer hyperlinks the selection


rfgamaral:
With #411 we fixed Markdown hyperlink pasting which stopped working, however, that introduced this new issue where pasting a URL on a text selection no longer hyperlinks the selection. Instead, the text selection is completely replaced with the URL. This mechanism came from Tiptap, and with the recent updates, the mechanism was moved to the [`pasteHandler` plugin](https://github.com/ueberdosis/tiptap/blob/main/packages/extension-link/src/link.ts#L172-L178), which got disabled with #411.

The fix for this issue involves reverting #411, and fixing the issue that #411 aimed to fix with a different solution.

One possible solution that I thought of was to implement the [`validate`](https://linkify.js.org/docs/options.html#validate) function from Linkify [here](https://github.com/ueberdosis/tiptap/blob/main/packages/extension-link/src/helpers/pasteHandler.ts#L81), and test for the Markdown hyperlink syntax. If the Markdown hyperlink syntax is found, return `false`, otherwise return `true`. This fix will need to be pushed upstream to Tiptap. Also, it may be worth it to introduce a new option to the `Link` extension to allows users control over this behaviour.

However, the above solution does not work because there's a bug in the Linkify `find` function, which I reported [here](https://github.com/Hypercontext/linkifyjs/issues/450). Once that's fixed, and a new Linkify version is released, we can then start working on pushing a fix to Tiptap.

PR #435, opened 2023-09-08
fix: Hyperlink a text selection when pasting a valid URL

rfgamaral:
## Overview

This PR closes https://github.com/Doist/typist/issues/423 with an implementation that does not require an upstream fix, one that increases the priority for our custom "paste extensions", making sure they are above the built-in `Link` extension. This will make sure that our handlers kick-in first, handling Markdown if and when appropriate. On top of that, it will check if the content being pasted is a valid URL, and if it is, let the built-in handlers do their thing (thus fixing https://github.com/Doist/typist/issues/423).

This approach will fix most of our sides - but not all - with a more consistent and predicable approach, which will make it easier to debug other issues that we have or might end up having in the future.

> **Note**
> For the detection of valid URLs I'm using the [`linkifyjs`](https://linkify.js.org/) library, something that Tiptap already uses internally, so we are not increasing the bundle. I'm just making it an explicit peer dependency, so that it can also be used in our products without dependency issues.

## PR Checklist

## Test plan

- Type a simple Markdown link (i.e. `[some text](https://example.com)`
- Type a Markdown link with title (i.e. `[some text](https://example.com "EXAMPLE")`
- Copy `[some text](https://example.com)` and paste into the editor (with `Ctrl/Cmd + Shift + V`)
- Copy `[some text](https://example.com "EXAMPLE")` and paste into the editor (with `Ctrl/Cmd + Shift + V`)
- Copy `[message link](message://%3cDM6PR05MB4777270E5C7A734F56BC519A9BE0A@DM6PR05MB4777.example.com%3e)` and paste into the editor (with `Ctrl/Cmd + Shift + V`)
- Type some text, make sure it's not hyperlinked, select that text, and paste a URL on the selection (try multiple URLs, like valid public ones, localhost ones, `message://` ones, simple ones like `example.com`)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -107,6 +107,7 @@
                 "@react-hookz/web": "^14.2.3 || >=15.x",
                 "emoji-regex": "^10.2.1",
                 "hast-util-is-element": "^2.1.0",
+                "linkifyjs": "^4.1.1",
                 "lodash-es": "^4.17.21",
                 "mdast-util-gfm-autolink-literal": "^1.0.0",
                 "mdast-util-gfm-strikethrough": "^1.0.0",
@@ -17479,9 +17480,9 @@
             }
         },
         "node_modules/linkifyjs": {
-            "version": "4.1.0",
-            "resolved": "https://registry.npmjs.org/linkifyjs/-/linkifyjs-4.1.0.tgz",
-            "integrity": "sha512-Ffv8VoY3+ixI1b3aZ3O+jM6x17cOsgwfB1Wq7pkytbo1WlyRp6ZO0YDMqiWT/gQPY/CmtiGuKfzDIVqxh1aCTA=="
+            "version": "4.1.1",
+            "resolved": "https://registry.npmjs.org/linkifyjs/-/linkifyjs-4.1.1.tgz",
+            "integrity": "sha512-zFN/CTVmbcVef+WaDXT63dNzzkfRBKT1j464NJQkV7iSgJU0sLBus9W0HBwnXK13/hf168pbrx/V/bjEHOXNHA=="
         },
         "node_modules/lint-staged": {
             "version": "14.0.1",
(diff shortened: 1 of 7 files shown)


gnapse (commented):
> Type a Markdown link with title (i.e. `[dolor sit amet](https://example.com "EXAMPLE"`)

Wow, I did not know about that syntax to add a title.

> (with `Ctrl/Cmd + Shift + V`)

Pressing this key combination did not paste anything for me. I was using Safari on macOS, which is apparently related because I then tried in Chrome, and it worked. Is this a known issue that this feature is not supported in Safari?

I marked all the related test plan items as checked, because it worked in Chrome. But it did not work in Safari.

> Type `vehicula congue`, make sure it's not hyperlinked, select that text, and paste a URL on the selection (try multiple URLs, like valid public ones, localhost ones, `message://` ones, simple ones like `example.com`)

In this part of the test plan, it worked as expected, except when using the `message://` url pattern.

https://github.com/Doist/typist/assets/15199/73b18c0d-1478-4231-9792-9f2a3cea5a5d

rfgamaral:
> Is this a known issue that this feature is not supported in Safari?

I honestly have no idea, maybe the Safari shortcuts [are different](https://support.apple.com/en-ca/HT201236):

image

Could you try those?

> except when using the `message://` url pattern.

Hum, well, I guess linkify does not detect these patterns as "valid URLs", based on the [docs](https://linkify.js.org/docs/#example-3-check-whether-a-string-is-a-valid-link) it seems that only works for URLs and e-mail addresses. We could probably rewrite the validation using [`find`](https://linkify.js.org/docs/linkifyjs.html#linkifyfind-str--type--null--opts--null) (with the `validation`option) instead of [`test`](https://linkify.js.org/docs/linkifyjs.html#linkifytest-str--type), but would be more complex to implement, and it's probably not worth it until we actually get a use case or an issue about this. I'll remove that `message://` URL pattern from the test plan.

gnapse (approved):
> Could you try those?

`Option-Shift-Command-V` worked

> I'll remove that `message://` URL pattern from the test plan.

With that removed, I can check the plan fully.

rfgamaral:
Thank you for this review, @gnapse 👍

PR #438, opened 2023-09-12
feat: Add `createEmptyParagraphEnd` command

rfgamaral:
## Overview

Adds a new command, `createEmptyParagraphEnd`, which creates an empty paragraph always at the end of the document.

This command that can be invoked like this:

- `editor.chain().focus().createEmptyParagraphEnd().run()`

This is just a helpful command that doesn't come built-in with Tiptap.

## PR Checklist

## Test plan

- Open the `Rich-text → Functions → Commands` story
- Type a multiline document (or copy and paste something)
- Place the text cursor in multiple locations, and in between press the `createEmptyParagraphEnd` button
    - [ ] Observe that regardless of the text cursor position, an empty paragraph is always created at the end of the document
- Open the `Plain-text → Functions → Commands` story
- Type a multiline document (or copy and paste something)
- Place the text cursor in multiple locations, and in between press the `createEmptyParagraphEnd` button
    - [ ] Observe that regardless of the text cursor position, an empty paragraph is always created at the end of the document


frankieyan:
Do we have a use case for this command?

PR #961, opened 2024-11-25
chore: Upgrade to `actions/cache@v4`

deorus:
It moves away from unsupported nodejs version to `node20`

See https://github.com/actions/cache?tab=readme-ov-file#v4

For https://github.com/Doist/platform-backlog/issues/628


rfgamaral (commented):
✅ 🙏

PR #1340, opened 2026-05-22
test(serializers): improve coverage and fix plain-text suggestion regex

rfgamaral:
## Overview

Fills in test coverage gaps for both the HTML and Markdown serializers, and fixes a small bug in the plain-text HTML serializer along the way.

**HTML serializer** (`src/serializers/html/html.test.ts`):
- New tests for code blocks with a language identifier, partially disabled extensions (coupled `attention`/`list` micromark constructs), multiple suggestion types in the same input, empty string input, and individual escape-disabling combinations.
- Fixes a bug where the plain-text serializer's suggestion regex only matched numeric IDs (`\d+`), while the rich-text path matched alphanumeric IDs (`\S+`). Both paths now use `\S+`, so suggestions like `[Henning M](mention://user:190200@doist.dev)` serialize correctly in plain-text mode too.

**Markdown serializer** (`src/serializers/markdown/markdown.test.ts`):
- New tests for fenced code blocks with a language identifier, Data URL images (replace base64 content with `NOT_SUPPORTED`), images with empty/missing `src`, ordered lists with the `start` attribute on empty items, plain-text whitespace preservation (the nbsp round-trip), alphanumeric suggestion IDs, multiple suggestion types in the same input, and empty string input.
- Fixes the plain-text suggestion tests that were using `RichTextKit` instead of `PlainTextKit` (so they weren't actually exercising the plain-text code path).
- Re-enables the previously skipped "only the paragraph rule is overwritten" test. It was set up to spy on `Turndown.prototype` after the serializer was already constructed, so the spies never captured anything. Spies are now set up at the `describe` level with `mockClear()` in `beforeEach`.

No behavior change outside of the plain-text suggestion regex fix.

## PR Checklist

## Test plan

- CI checks should be green

### DIFF
--- a/src/serializers/html/html.test.ts
+++ b/src/serializers/html/html.test.ts
@@ -408,6 +408,18 @@ Answer: [Doist Frontend](channel://190200)`),
                 )
             })
 
+            test('suggestion inside parentheses does not absorb the trailing parenthesis', () => {
+                const htmlSerializer = createHTMLSerializer(
+                    getSchema([PlainTextKit, createSuggestionExtension('mention')]),
+                )
+
+                expect(
+                    htmlSerializer.serialize('([Henning M](mention://user:190200@doist.dev))'),
+                ).toBe(
+                    '<p>(<span data-mention data-id="user:190200@doist.dev" data-label="Henning M"></span>)</p>',
+                )
+            })
+
             test('multiple suggestion types in the same input are serialized', () => {
                 const htmlSerializer = createHTMLSerializer(
                     getSchema([
--- a/src/serializers/html/html.test.ts
+++ b/src/serializers/html/html.test.ts
@@ -200,6 +200,10 @@ const MARKDOWN_INPUT_CODE_BLOCK = `\`\`\`
 </html>
 \`\`\``
 
+const MARKDOWN_INPUT_CODE_BLOCK_WITH_LANGUAGE = `\`\`\`javascript
+console.log("hello")
+\`\`\``
+
 const MARKDOWN_INPUT_INDENTED_BLOCK_ELEMENTS = `1. Blockquote:
     > Dorothy followed her through many of the beautiful rooms in her castle.
 2. Image:
@@ -256,6 +260,10 @@ describe('HTML Serializer', () => {
 
             beforeEach(() => {
                 htmlSerializer = createHTMLSerializer(getSchema([PlainTextKit]))
+            })
+
+            test('empty string returns empty output', () => {
+                expect(htmlSerializer.serialize('')).toBe('')
             })
 
             test('special ASCII characters are converted to HTML entities', () => {
@@ -330,6 +338,12 @@ describe('HTML Serializer', () => {
                 )
             })
 
+            test('code block with language syntax is preserved', () => {
+                expect(htmlSerializer.serialize(MARKDOWN_INPUT_CODE_BLOCK_WITH_LANGUAGE)).toBe(
+                    '<p>```javascript</p><p>console.log(&quot;hello&quot;)</p><p>```</p>',
+                )
+            })
+
             test('indented block elements syntax is preserved', () => {
                 expect(htmlSerializer.serialize(MARKDOWN_INPUT_INDENTED_BLOCK_ELEMENTS)).toBe(
                     '<p>1. Blockquote:</p><p>    &gt; Dorothy followed her through many of the beautiful rooms in her castle.</p><p>2. Image:</p><p>    ![Octobi Wan Catnobi](https://octodex.github.com/images/octobiwan.jpg)</p><p>3. Codeblock:</p><p>    ```</p><p>    &lt;html&gt;</p><p>      &lt;head&gt;</p><p>        &lt;title&gt;Test&lt;/title&gt;</p><p>      &lt;/head&gt;</p><p>    &lt;/html&gt;</p><p>    ```</p>',
@@ -381,6 +395,36 @@ Answer: [Doist Frontend](channel://190200)`),
                     '<p>Question: What&#39;s the best channel on Twist?</p><p>Answer: <span data-channel data-id="190200" data-label="Doist Frontend"></span></p>',
                 )
             })
+
+            test('suggestion with alphanumeric ID is serialized', () => {
+                const htmlSerializer = createHTMLSerializer(
+                    getSchema([PlainTextKit, createSuggestionExtension('mention')]),
+                )
+
+                expect(
+                    htmlSerializer.serialize('[Henning M](mention://user:190200@doist.dev)'),
+                ).toBe(
+                    '<p><span data-mention data-id="user:190200@doist.dev" data-label="Henning M"></span></p>',
+                )
+            })
+
+            test('multiple suggestion types in the same input are serialized', () => {
+                const htmlSerializer = createHTMLSerializer(
+                    getSchema([
+                        PlainTextKit,
+                        createSuggestionExtension('mention'),
+                        createSuggestionExtension('channel'),
+                    ]),
+                )
+
+                expect(
+                    htmlSerializer.serialize(
+                        'Hey [Alice](mention://123) and [Bob](mention://456), check [General](channel://789)',
+                    ),
+                ).toBe(
+                    '<p>Hey <span data-mention data-id="123" data-label="Alice"></span> and <span data-mention data-id="456" data-label="Bob"></span>, check <span data-channel data-id="789" data-label="General"></span></p>',
+                )
+            })
         })
     })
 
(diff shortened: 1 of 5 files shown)


src/serializers/html/html.ts
```diff
@@ -61,7 +61,7 @@ function createHTMLSerializerForPlainTextEditor(schema: Schema) {
                     const linkSchema = kebabCase(suggestionNode.name.replace(/Suggestion$/, ''))
 
                     htmlResult = htmlResult.replace(
-                        new RegExp(`\\[([^\\[]+)\\]\\((?:${linkSchema}):\\/\\/(\\d+)\\)`, 'gm'),
+                        new RegExp(`\\[([^\\[]+)\\]\\((?:${linkSchema}):\\/\\/(\\S+)\\)`, 'gm'),
```
doistbot: [P1] This `\S+` capture now eats an extra `)` when a suggestion is immediately followed by another closing paren, so plain-text serialization corrupts IDs in cases like `([Alice](mention://user:190200@doist.dev))` by producing `data-id="user:190200@doist.dev)"`. Match up to the link's closing paren instead of any non-whitespace, for example with `([^\s)]+)` or another non-greedy pattern, and add a regression test for the parenthesized case.

src/serializers/html/html.test.ts
```diff
@@ -381,6 +395,36 @@ Answer: [Doist Frontend](channel://190200)`),
...
+
+                expect(
+                    htmlSerializer.serialize('[Henning M](mention://user:190200@doist.dev)'),
+                ).toBe(
+                    '<p><span data-mention data-id="user:190200@doist.dev" data-label="Henning M"></span></p>',
+                )
+            })
+
+            test('multiple suggestion types in the same input are serialized', () => {
+                const htmlSerializer = createHTMLSerializer(
```
doistbot, line 424: [P3] This setup duplicates the `createHTMLSerializer` initialization from the tests above it. To reduce repetition, consider registering both the `mention` and `channel` extensions in a single `beforeEach` block for this `describe` scope, mirroring the pattern used for the rich-text suggestion tests (around line 813).

src/serializers/markdown/markdown.test.ts
```diff
@@ -371,6 +388,36 @@ Answer: [Henning M](mention://963827)`)
...
+
+                expect(
+                    markdownSerializer.serialize(
+                        '<p><span data-mention data-id="user:190200@doist.dev" data-label="Henning M">@Henning M</span></p>',
+                    ),
+                ).toBe('[Henning M](mention://user:190200@doist.dev)')
+            })
+
+            test('multiple suggestion types in the same input are serialized', () => {
+                const markdownSerializer = createMarkdownSerializer(
```
doistbot, line 385: [P3] Similar to the HTML serializer tests, this duplicates the `createMarkdownSerializer` initialization. You can consolidate the setup by moving the schema creation with both `mention` and `channel` extensions into a shared `beforeEach` for this block.

src/serializers/html/html.ts
doistbot: [P2] This is still a second hand-written parser for suggestion links. `rehypeSuggestions` already parses the same `scheme://id` format, and the old `\d+`/`\S+` mismatch happened because these paths drifted. Please extract the suggestion-link regex/parsing into a shared helper and use it here too, so both serializer paths stay aligned.

src/serializers/markdown/markdown.test.ts
```diff
@@ -287,15 +304,15 @@ I think I'll use it to format all of my documents from now on.`,
...
-                const addRuleMock = vi.spyOn(Turndown.prototype, 'addRule')
+            test('leading whitespace is preserved', () => {
+                expect(markdownSerializer.serialize('<p>1. First</p><p>    1. Indented</p>'))
+                    .toBe(`1. First
+    1. Indented`)
+            })
 
-                expect(useMock).toHaveBeenCalledTimes(1)
-                expect(addRuleMock).toHaveBeenLastCalledWith('paragraph', {
+            test('only the paragraph rule is overwritten', () => {
```
doistbot: [P2] This test asserts Turndown internals (`use`/`addRule` on the prototype) rather than serialized output. That makes the suite brittle to harmless refactors and adds global test state just to verify implementation details. Please replace it with behavior-based coverage or drop it.

src/serializers/markdown/markdown.test.ts
```diff
@@ -581,6 +644,18 @@ Octobi Wan Catnobi: ![](https://octodex.github.com/images/octobiwan.jpg) - These
...
 
+            test('images with Data URLs replace base64 content with NOT_SUPPORTED', () => {
+                expect(
+                    markdownSerializer.serialize(
+                        '<p><img src="data:image/png;base64,iVBORw0KGgo=" alt="screenshot"></p>',
+                    ),
+                ).toBe('![screenshot](data:image/png;base64,NOT_SUPPORTED)')
+            })
+
+            test('images with empty src are omitted', () => {
```
doistbot, line 635: [P2] This adds coverage for `src=""`, but the adjacent missing-`src` case is still untested. That matters here because `String(node.getAttribute('src'))` turns a missing attribute into `'null'`, so `test` currently serializes differently instead of being omitted. Add a missing-`src` example next to this one.

doistbot (commented):
This PR significantly improves test coverage across both the HTML and Markdown serializers while fixing a plain-text serialization bug for non-numeric suggestion IDs. These thoughtful additions create a much stronger safety net against regressions and ensure better consistency between the plain-text and rich-text pathways. There are a few remaining details to refine, including addressing an edge case where the new regex captures trailing parentheses, extracting suggestion parsing into a shared helper, shifting a Turndown internals test to behavior-based coverage, handling missing image sources, and consolidating duplicated test setups.

<sub>[Share Feedback](https://github.com/Doist/doistbot-app/issues/new?template=review-feedback.yml&title=Review+feedback+for+Doist%2Ftypist&pr_url=https%3A%2F%2Fgithub.com%2FDoist%2Ftypist%2Fpull%2F1340&datadog_url=https%3A%2F%2Fapp.datadoghq.com%2Flogs%3Fquery%3Dservice%253A%2528doistbot-app-sandbox%2520OR%2520doistbot-app%2529%2520%2540meta.deliveryId%253Abade7db0-5603-11f1-8709-990223f8b2c5%26agg_m%3Dcount%26agg_m_source%3Dbase%26agg_t%3Dcount%26clustering_pattern_field_path%3Dmessage%26cols%3Dhost%252Cservice%26fromUser%3Dtrue%26messageDisplay%3Dinline%26refresh_mode%3Dsliding%26storage%3Dhot%26stream_sort%3Ddesc%26viz%3Dstream%26live%3Dtrue) • [Review Logs](https://app.datadoghq.com/logs?query=service%3A%28doistbot-app-sandbox%20OR%20doistbot-app%29%20%40meta.deliveryId%3Abade7db0-5603-11f1-8709-990223f8b2c5&agg_m=count&agg_m_source=base&agg_t=count&clustering_pattern_field_path=message&cols=host%2Cservice&fromUser=true&messageDisplay=inline&refresh_mode=sliding&storage=hot&stream_sort=desc&viz=stream&live=true)</sub>

src/serializers/html/html.test.ts
rfgamaral: These tests intentionally register only the suggestion extension under test (mention-only or channel-only) to verify each works in isolation. Consolidating into a shared `beforeEach` with both extensions registered would make those tests redundant with the existing `multiple suggestion types` test.

src/serializers/markdown/markdown.test.ts
rfgamaral: These tests intentionally register only the suggestion extension under test (mention-only or channel-only) to verify each works in isolation. Consolidating into a shared `beforeEach` with both extensions registered would make those tests redundant with the existing `multiple suggestion types` test.

src/serializers/markdown/markdown.test.ts
rfgamaral: Dropped the test. The intent (plain-text mode doesn't apply transformations beyond paragraph handling) is already covered by the surrounding behavior tests, and the prototype-spy approach was tied to Turndown specifically.

src/serializers/html/html.ts
rfgamaral: Pushing back on this. The two regexes aren't parsing the same thing:

- The plain-text path scans raw Markdown text and has to match the full `[label](scheme://id)` syntax, with the ID terminating before the link's closing `)`. That's why it uses `[^\s)]+`.
- The rich-text path operates on an already-parsed href attribute, so the ID has no surrounding Markdown wrapper and trailing parens aren't part of it. `\S+` is correct there.

The character class for the ID legitimately differs between the two contexts. Extracting a shared piece would either force one path to use a less-fitting pattern, or end up as a constant so trivial that it doesn't materially prevent drift.

The original `\d+` / `\S+` mismatch was a real issue (alphanumeric IDs don't work in plain-text), but it sat untouched for years because no one hit it in production. It's now fixed and covered by a regression test. I don't think the abstraction pays for itself.

PR #1341, opened 2026-05-22
fix(serializers): include trigger character in suggestion HTML output

rfgamaral:
## Overview

Makes the HTML serializer output directly renderable for suggestion nodes (mentions, channels, etc.) by including the trigger character and label as text content inside the span, instead of leaving it empty.

Before, the serializer output empty spans like `<span data-mention data-id="123" data-label="Alice"></span>`. This only "worked" because Tiptap parses the data attributes and re-renders the visible content via the suggestion extension's `renderHTML`. If you drop this HTML into any other rendering context (a browser, a React component using `dangerouslySetInnerHTML`, etc.), the mention shows as nothing because the span is empty.

After this PR, the output is `<span data-mention data-id="123" data-label="Alice">@Alice</span>` - self-contained and directly renderable. Tiptap still understands it (it reads from `data-label`), and any other consumer can render it as-is.

The trigger character is configurable per suggestion extension (`@`, `#`, etc.), so `createSuggestionExtension` now exposes it on the node spec via `extendNodeSchema`. The `rehype-suggestions` plugin reads it from the schema and prepends it to the label text. Default trigger char (`@`) lives in a new shared constant under `src/constants/suggestions.ts`.

This is forward-looking: it unblocks future use cases where the HTML serializer's output is rendered without going through the editor (for example, replacing `MarkdownContent` in todoist-web with a Markdown-to-React renderer built on these serializers). It also gives consumers a direct way to read the trigger character from a specific extension instance via `schema.nodes.{type}Suggestion.spec.triggerChar`, without having to track it separately or hardcode it.

## PR Checklist

## Test plan

- Smoke test the `Rich-text → Default` story in Storybook
    - [x] Type `@` and select a mention from the dropdown - it still renders and serializes the same
    - [x] Type `#` and select a channel suggestion - same
- CI checks should be green

### DIFF
--- a/src/helpers/serializer.ts
+++ b/src/helpers/serializer.ts
@@ -1,17 +1,36 @@
 import { kebabCase } from 'lodash-es'
 
+import { DEFAULT_SUGGESTION_TRIGGER_CHAR } from '../constants/suggestions'
+
 import type { ParseRule, Schema } from '@tiptap/pm/model'
 
 /**
- * Builds a partial regular expression that includes valid URL schemas used by all the available
- * suggestion nodes from the given editor schema.
+ * Information derived from the suggestion nodes available in the editor schema, used by the
+ * HTML serializer to identify and transform suggestion links into spans.
+ */
+type SuggestionSchemaInfo = {
+    /**
+     * A partial regular expression that matches the URL schemes used by all the available
+     * suggestion nodes (e.g. `(?:mention|channel)://`).
+     */
+    urlSchemeRegex: string
+
+    /**
+     * A map from each URL scheme (e.g. `mention`, `channel`) to its configured trigger character
+     * (e.g. `@`, `#`).
+     */
+    triggerCharByScheme: Map<string, string>
+}
+
+/**
+ * Builds the information derived from all the suggestion nodes available in the given editor
+ * schema, in a single iteration. Returns `null` if there are no suggestion nodes in the schema.
  *
  * @param schema The editor schema to be used for suggestion nodes detection.
  *
- * @returns A partial regular expression with valid URL schemas for the available suggestion nodes,
- * `null` if there are no suggestion nodes in the editor schema.
+ * @returns A `SuggestionSchemaInfo` object, or `null` if there are no suggestion nodes.
  */
-function buildSuggestionSchemaPartialRegex(schema: Schema) {
+function buildSuggestionSchemaInfo(schema: Schema): SuggestionSchemaInfo | null {
     const suggestionNodes = Object.values(schema.nodes).filter((node) =>
         node.name.endsWith('Suggestion'),
     )
@@ -20,9 +39,22 @@ function buildSuggestionSchemaPartialRegex(schema: Schema) {
         return null
     }
 
-    return `(?:${suggestionNodes
-        .map((suggestionNode) => kebabCase(suggestionNode.name.replace(/Suggestion$/, '')))
-        .join('|')})://`
+    const triggerCharByScheme = new Map(
+        suggestionNodes.map((node) => [
+            kebabCase(node.name.replace(/Suggestion$/, '')),
+            String(
+                (node.spec as { triggerChar?: string }).triggerChar ??
+                    DEFAULT_SUGGESTION_TRIGGER_CHAR,
+            ),
+        ]),
+    )
+
+    const urlSchemes = [...triggerCharByScheme.keys()]
+
+    return {
+        urlSchemeRegex: `(?:${urlSchemes.join('|')})://`,
+        triggerCharByScheme,
+    }
 }
 
 /**
@@ -44,4 +76,4 @@ function extractTagsFromParseRules(
         .map((rule) => rule.tag as keyof HTMLElementTagNameMap)
 }
 
-export { buildSuggestionSchemaPartialRegex, extractTagsFromParseRules }
+export { buildSuggestionSchemaInfo, extractTagsFromParseRules }
--- a/src/serializers/html/html.test.ts
+++ b/src/serializers/html/html.test.ts
@@ -837,7 +837,7 @@ See the section on [\`code\`](#code).</p>`)
                     htmlSerializer.serialize(`Question: Who's the head of the Frontend team?
 Answer: [Henning M](mention://user:190200@doist.dev)`),
                 ).toBe(
-                    '<p>Question: Who\'s the head of the Frontend team?<br>Answer: <span data-mention="" data-id="user:190200@doist.dev" data-label="Henning M"></span></p>',
+                    '<p>Question: Who\'s the head of the Frontend team?<br>Answer: <span data-mention="" data-id="user:190200@doist.dev" data-label="Henning M">@Henning M</span></p>',
                 )
             })
 
@@ -846,7 +846,7 @@ Answer: [Henning M](mention://user:190200@doist.dev)`),
                     htmlSerializer.serialize(`Question: Who's the head of the Frontend team?
 Answer: [Henning M](mention://963827)`),
                 ).toBe(
-                    '<p>Question: Who\'s the head of the Frontend team?<br>Answer: <span data-mention="" data-id="963827" data-label="Henning M"></span></p>',
+                    '<p>Question: Who\'s the head of the Frontend team?<br>Answer: <span data-mention="" data-id="963827" data-label="Henning M">@Henning M</span></p>',
                 )
             })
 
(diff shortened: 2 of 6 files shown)


src/serializers/html/plugins/rehype-suggestions.ts
```diff
@@ -31,15 +31,25 @@ function rehypeSuggestions(schema: Schema): Transformer {
                 const [, urlScheme, id] =
                     /^([a-z-]+):\/\/(\S+)$/i.exec(String(node.properties?.href)) || []
 
-                // Replace the link element with a span containing the suggestion attributes
+                // Replace the link element with a span containing the suggestion attributes,
```
doistbot, line 34: [P1] This correctly updates the rich-text serialization path, but the plain-text serialization branch in `src/serializers/html/html.ts` (which uses a regex replacement instead of `rehype-suggestions`) still outputs empty spans. Consider updating the `htmlResult.replace` logic for `isPlainTextDocument` to also include the trigger character and label, ensuring consistency across both modes.

src/helpers/serializer.ts
```diff
@@ -20,9 +39,22 @@ function buildSuggestionSchemaPartialRegex(schema: Schema) {
...
     }
 
-    return `(?:${suggestionNodes
-        .map((suggestionNode) => kebabCase(suggestionNode.name.replace(/Suggestion$/, '')))
-        .join('|')})://`
+    const triggerCharByScheme = new Map(
+        suggestionNodes.map((node) => [
+            kebabCase(node.name.replace(/Suggestion$/, '')),
+            String(
+                (node.spec as { triggerChar?: string }).triggerChar ??
```
doistbot, line 46: [P1] This makes HTML serialization depend on `node.spec.triggerChar`, but `getHTMLSerializerInstance()` still caches by `computeSchemaId(schema)`, which only includes mark/node names. Two schemas with the same suggestion nodes but different `triggerChar` values will now share one serializer, and the later one will emit the first schema's prefix. Include the trigger chars in the cache key, or don't reuse serializer instances across differently configured schemas.

src/helpers/serializer.ts
```diff
@@ -20,9 +39,22 @@ function buildSuggestionSchemaPartialRegex(schema: Schema) {
         return null
     }
 
-    return `(?:${suggestionNodes
-        .map((suggestionNode) => kebabCase(suggestionNode.name.replace(/Suggestion$/, '')))
-        .join('|')})://`
+    const triggerCharByScheme = new Map(
+        suggestionNodes.map((node) => [
+            kebabCase(node.name.replace(/Suggestion$/, '')),
```
doistbot, line 44: [P3] The scheme name extraction `kebabCase(node.name.replace(/Suggestion$/, ''))` is duplicated here, in `src/serializers/html/html.ts`, and in `src/serializers/markdown/plugins/suggestion.ts`.

Consider extracting this into a shared helper function (e.g., `getSuggestionScheme(nodeName)`) to remove the repeated string manipulation across the codebase.

src/helpers/serializer.ts
```diff
@@ -1,17 +1,36 @@
...
+
 import type { ParseRule, Schema } from '@tiptap/pm/model'
 
 /**
- * Builds a partial regular expression that includes valid URL schemas used by all the available
- * suggestion nodes from the given editor schema.
+ * Information derived from the suggestion nodes available in the editor schema, used by the
+ * HTML serializer to identify and transform suggestion links into spans.
+ */
+type SuggestionSchemaInfo = {
```
doistbot, line 11: [P3] Per `docs/typescript-guidelines/types.md`, use an `interface` instead of a `type` alias for object shapes.

src/serializers/html/plugins/rehype-suggestions.ts
```diff
@@ -31,15 +31,25 @@ function rehypeSuggestions(schema: Schema): Transformer {
...
+                // keeping the visible label (prefixed with the trigger character) as text content
+                // so the span renders correctly when used outside of an editor.
                 if (urlScheme && id && isHastTextNode(node.children[0])) {
+                    const label = node.children[0].value
+
+                    // The URL scheme was matched against the regex built from the same map of
+                    // suggestion nodes, so the trigger character is guaranteed to exist
+                    const triggerChar = suggestionSchemaInfo.triggerCharByScheme.get(
```
doistbot, line 42: [P2] Instead of bypassing type safety with `as string` and leaving a justification comment, consider narrowing the type with a standard guard (e.g., `if (triggerChar === undefined) return`). This adheres to the TypeScript philosophy of runtime type narrowing over assertions.

src/serializers/html/plugins/rehype-suggestions.ts
```diff
@@ -13,15 +13,15 @@ import type { Transformer } from 'unified'
...
 
     // Return the tree as-is if the editor does not support suggestions
-    if (!suggestionSchemaPartialRegex) {
+    if (!suggestionSchemaInfo) {
         return (tree: HastNode) => tree
     }
 
     return (...[tree]: Parameters<Transformer>): ReturnType<Transformer> => {
-        const suggestionSchemaRegex = new RegExp(`^${suggestionSchemaPartialRegex}`)
+        const suggestionSchemaRegex = new RegExp(`^${suggestionSchemaInfo.urlSchemeRegex}`)
```
doistbot, line 24: [P2] Compiling this regular expression inside the returned transformer means it gets recompiled on every single AST transformation. Move it outside the closure (above line 23) so it only compiles once when the plugin initializes.

src/serializers/html/html.test.ts
```diff
@@ -855,7 +855,7 @@ Answer: [Henning M](mention://963827)`),
                     htmlSerializer.serialize(`Question: What's the best channel on Twist?
 Answer: [Doist Frontend](channel://190200)`),
                 ).toBe(
-                    '<p>Question: What\'s the best channel on Twist?<br>Answer: <span data-channel="" data-id="190200" data-label="Doist Frontend"></span></p>',
+                    '<p>Question: What\'s the best channel on Twist?<br>Answer: <span data-channel="" data-id="190200" data-label="Doist Frontend">@Doist Frontend</span></p>',
```
doistbot, line 858: [P2] This assertion tests the default fallback (`@Doist Frontend`), leaving the PR's core feature—rendering custom trigger characters—untested at the serializer level. Update the suite's `beforeEach` block to configure the channel suggestion with `#` (`createSuggestionExtension('channel').configure({ triggerChar: '#' })`) and assert `#Doist Frontend` here. This proves the custom configuration properly flows through the entire HTML serialization pipeline.

doistbot (commented):
This PR updates the HTML serializer to include the trigger character and label directly within suggestion node spans, ensuring the output is self-contained and renderable outside of Tiptap. This is a great forward-looking improvement that will significantly enhance the flexibility of the serializer output across different rendering contexts. A few adjustments are needed to address caching edge cases with custom trigger characters, ensure consistency in the plain-text serialization branch, and refine testing, along with some minor TypeScript and performance optimizations.

<sub>[Share Feedback](https://github.com/Doist/doistbot-app/issues/new?template=review-feedback.yml&title=Review+feedback+for+Doist%2Ftypist&pr_url=https%3A%2F%2Fgithub.com%2FDoist%2Ftypist%2Fpull%2F1341&datadog_url=https%3A%2F%2Fapp.datadoghq.com%2Flogs%3Fquery%3Dservice%253A%2528doistbot-app-sandbox%2520OR%2520doistbot-app%2529%2520%2540meta.deliveryId%253A5e243850-5615-11f1-9bf4-77f26cf50b1a%26agg_m%3Dcount%26agg_m_source%3Dbase%26agg_t%3Dcount%26clustering_pattern_field_path%3Dmessage%26cols%3Dhost%252Cservice%26fromUser%3Dtrue%26messageDisplay%3Dinline%26refresh_mode%3Dsliding%26storage%3Dhot%26stream_sort%3Ddesc%26viz%3Dstream%26live%3Dtrue) • [Review Logs](https://app.datadoghq.com/logs?query=service%3A%28doistbot-app-sandbox%20OR%20doistbot-app%29%20%40meta.deliveryId%3A5e243850-5615-11f1-9bf4-77f26cf50b1a&agg_m=count&agg_m_source=base&agg_t=count&clustering_pattern_field_path=message&cols=host%2Cservice&fromUser=true&messageDisplay=inline&refresh_mode=sliding&storage=hot&stream_sort=desc&viz=stream&live=true)</sub>

src/helpers/serializer.ts
rfgamaral: Pushing back. Per the discussion in https://github.com/Doist/shared-configs/pull/39#discussion_r2883602718, the convention is to use `type` for object shapes by default and only reach for `interface` in specific cases (declaration merging, public extensible APIs). `SuggestionSchemaInfo` is a plain internal object shape with neither of those needs, so `type` is the right choice here.

src/serializers/html/plugins/rehype-suggestions.ts
rfgamaral: The cast is safe by construction: `urlScheme` comes from a regex match using the same regex that built the map. Every scheme added to the regex was added to the map in the same iteration, with identical keys. The lookup can't miss.

Adding a runtime guard would be dead code (it can never fire) and would also lie to the reader by implying `triggerChar` might be undefined when it can't be. I'd rather keep the cast and the comment that documents the invariant.

pedroalves0 (commented):
✅

PR #1342, opened 2026-05-23
fix(serializers): follow-up fixes to suggestion HTML output

rfgamaral:
## Overview

Follow-up fixes to [#1341](https://github.com/Doist/typist/pull/1341), addressing review feedback that wasn't pushed before that PR was merged. The fixes cover both real defects and small refactors:

- **Plain-text suggestion path was still emitting empty spans.** The original PR only updated the rich-text rehype plugin; the plain-text regex-replacement branch in `html.ts` was overlooked. The label is now included (prefixed with the configured trigger character) as text content in both modes.
- **Schema cache key didn't account for `triggerChar`.** `computeSchemaId` only included mark and node names, so two schemas with the same suggestion nodes but different `triggerChar` configs produced the same cache key and shared one serializer. This was a latent bug before, but #1341 made the output depend on `triggerChar`, so cache pollution would now render the wrong prefix.
- **Custom trigger character wasn't exercised in tests.** The rich-text and plain-text suggestion tests both used the default `@` for every node, so the trigger-char wiring wasn't covered end-to-end. The channel suggestion test now configures `triggerChar: '#'` and asserts `#Doist Frontend`.
- **Suggestion regex was recompiled on every AST transformation** in `rehype-suggestions`. Moved out of the returned closure so it compiles once per plugin invocation.
- **Suggestion URL scheme extraction was duplicated** (`kebabCase(node.name.replace(/Suggestion$/, ''))`) across three files. Extracted into a `getSuggestionUrlScheme` helper in `src/helpers/serializer.ts`.

## PR Checklist

## Test plan

- Smoke test the `Rich-text → Default` story in Storybook
    - [ ] Type `@` and select a mention from the dropdown - it still renders and serializes the same
    - [ ] Type `#` and select a channel suggestion - same
- CI checks should be green

### DIFF
--- a/src/helpers/schema.ts
+++ b/src/helpers/schema.ts
@@ -24,23 +24,13 @@ function isPlainTextDocument(schema: Schema): boolean {
 
 /**
  * Computes a string ID that identifies a given editor schema which can be used for object mapping.
- * The trigger character of each suggestion node is included in the ID so schemas that differ only
- * in suggestion configuration produce distinct cache keys.
  *
  * @param schema The current editor document schema.
  *
  * @returns A string ID matching the editor schema.
  */
 function computeSchemaId(schema: Schema) {
-    const suggestionTriggers = Object.values(schema.nodes)
-        .filter((node) => node.name.endsWith('Suggestion'))
-        .map((node) => `${node.name}=${(node.spec as { triggerChar?: string }).triggerChar ?? ''}`)
-
-    return [
-        ...Object.keys(schema.marks),
-        ...Object.keys(schema.nodes),
-        ...suggestionTriggers,
-    ].join()
+    return [...Object.keys(schema.marks), ...Object.keys(schema.nodes)].join()
 }
 
 export { computeSchemaId, isMultilineDocument, isPlainTextDocument }
--- a/src/helpers/schema.ts
+++ b/src/helpers/schema.ts
@@ -24,13 +24,23 @@ function isPlainTextDocument(schema: Schema): boolean {
 
 /**
  * Computes a string ID that identifies a given editor schema which can be used for object mapping.
+ * The trigger character of each suggestion node is included in the ID so schemas that differ only
+ * in suggestion configuration produce distinct cache keys.
  *
  * @param schema The current editor document schema.
  *
  * @returns A string ID matching the editor schema.
  */
 function computeSchemaId(schema: Schema) {
-    return [...Object.keys(schema.marks), ...Object.keys(schema.nodes)].join()
+    const suggestionTriggers = Object.values(schema.nodes)
+        .filter((node) => node.name.endsWith('Suggestion'))
+        .map((node) => `${node.name}=${(node.spec as { triggerChar?: string }).triggerChar ?? ''}`)
+
+    return [
+        ...Object.keys(schema.marks),
+        ...Object.keys(schema.nodes),
+        ...suggestionTriggers,
+    ].join()
 }
 
 export { computeSchemaId, isMultilineDocument, isPlainTextDocument }
--- a/src/serializers/html/html.test.ts
+++ b/src/serializers/html/html.test.ts
@@ -827,7 +827,7 @@ See the section on [\`code\`](#code).</p>`)
                     getSchema([
                         RichTextKit,
                         createSuggestionExtension('mention'),
-                        createSuggestionExtension('channel'),
+                        createSuggestionExtension('channel').configure({ triggerChar: '#' }),
                     ]),
                 )
             })
@@ -855,7 +855,7 @@ Answer: [Henning M](mention://963827)`),
                     htmlSerializer.serialize(`Question: What's the best channel on Twist?
 Answer: [Doist Frontend](channel://190200)`),
                 ).toBe(
-                    '<p>Question: What\'s the best channel on Twist?<br>Answer: <span data-channel="" data-id="190200" data-label="Doist Frontend">@Doist Frontend</span></p>',
+                    '<p>Question: What\'s the best channel on Twist?<br>Answer: <span data-channel="" data-id="190200" data-label="Doist Frontend">#Doist Frontend</span></p>',
                 )
             })
 
@@ -865,7 +865,7 @@ Answer: [Doist Frontend](channel://190200)`),
                         'Hey [Alice](mention://123) and [Bob](mention://456), check [General](channel://789)',
                     ),
                 ).toBe(
-                    '<p>Hey <span data-mention="" data-id="123" data-label="Alice">@Alice</span> and <span data-mention="" data-id="456" data-label="Bob">@Bob</span>, check <span data-channel="" data-id="789" data-label="General">@General</span></p>',
+                    '<p>Hey <span data-mention="" data-id="123" data-label="Alice">@Alice</span> and <span data-mention="" data-id="456" data-label="Bob">@Bob</span>, check <span data-channel="" data-id="789" data-label="General">#General</span></p>',
                 )
             })
         })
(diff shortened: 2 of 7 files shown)


src/serializers/html/html.ts
```diff
@@ -58,11 +60,16 @@ function createHTMLSerializerForPlainTextEditor(schema: Schema) {
...
+
+                    const triggerChar = String(
+                        (suggestionNode.spec as { triggerChar?: string }).triggerChar ??
+                            DEFAULT_SUGGESTION_TRIGGER_CHAR,
+                    )
 
                     htmlResult = htmlResult.replace(
                         new RegExp(`\\[([^\\[]+)\\]\\((?:${linkSchema}):\\/\\/([^\\s)]+)\\)`, 'gm'),
-                        `<span data-${linkSchema} data-id="$2" data-label="$1"></span>`,
+                        `<span data-${linkSchema} data-id="$2" data-label="$1">${triggerChar}$1</span>`,
```
doistbot: [P2] If `triggerChar` is `$` (e.g. for cashtags), this string replacement will break because `$$1` evaluates to the literal string `$1` instead of evaluating the capture group.

Consider passing a replacer function instead, which also gives you an opportunity to escape the trigger character in case it contains HTML-unsafe symbols:
```typescript
(_, label, id) => `<span data-${linkSchema} data-id="${id}" data-label="${label}">${escape(triggerChar)}${label}</span>`
```

src/helpers/schema.ts
```diff
@@ -24,13 +24,23 @@ function isPlainTextDocument(schema: Schema): boolean {
...
+ * The trigger character of each suggestion node is included in the ID so schemas that differ only
+ * in suggestion configuration produce distinct cache keys.
  *
  * @param schema The current editor document schema.
  *
  * @returns A string ID matching the editor schema.
  */
 function computeSchemaId(schema: Schema) {
-    return [...Object.keys(schema.marks), ...Object.keys(schema.nodes)].join()
+    const suggestionTriggers = Object.values(schema.nodes)
```
doistbot: [P2] The logic to find suggestion nodes (`Object.values(schema.nodes).filter(...)`) is now duplicated here, in `buildSuggestionSchemaInfo`, and in `createHTMLSerializerForPlainTextEditor`. Consider extracting a `getSuggestionNodes(schema)` helper to share this traversal.

src/serializers/html/html.ts
```diff
@@ -58,11 +60,16 @@ function createHTMLSerializerForPlainTextEditor(schema: Schema) {
             Object.values(schema.nodes)
                 .filter((node) => node.name.endsWith('Suggestion'))
                 .forEach((suggestionNode) => {
-                    const linkSchema = kebabCase(suggestionNode.name.replace(/Suggestion$/, ''))
+                    const linkSchema = getSuggestionUrlScheme(suggestionNode)
+
+                    const triggerChar = String(
```
doistbot: [P2] Similar to `getSuggestionUrlScheme`, the extraction of the trigger character is duplicated here, in `buildSuggestionSchemaInfo`, and in `computeSchemaId` (which inconsistently falls back to `''`). Consider extracting a `getSuggestionTriggerChar(node)` helper to centralize this and ensure `DEFAULT_SUGGESTION_TRIGGER_CHAR` is used consistently.

src/serializers/html/html.ts
```diff
@@ -58,11 +60,16 @@ function createHTMLSerializerForPlainTextEditor(schema: Schema) {
             Object.values(schema.nodes)
                 .filter((node) => node.name.endsWith('Suggestion'))
                 .forEach((suggestionNode) => {
-                    const linkSchema = kebabCase(suggestionNode.name.replace(/Suggestion$/, ''))
+                    const linkSchema = getSuggestionUrlScheme(suggestionNode)
```
doistbot: [P2] We already centralize suggestion-scheme and trigger-char discovery in `buildSuggestionSchemaInfo()`, but this plain-text branch re-scans `schema.nodes` and re-applies the same defaulting here. That leaves the rich-text and plain-text HTML serializers with separate metadata logic again—the same drift this PR is fixing. Reuse `buildSuggestionSchemaInfo()` here and iterate its `triggerCharByScheme` map instead.

src/helpers/schema.ts
doistbot: [P2] `computeSchemaId()` is shared by both the HTML and Markdown serializer caches, but `triggerChar` only affects HTML output. Encoding it into the global schema ID couples this helper to one serializer and forces unnecessary Markdown cache misses for schemas that would serialize identically. Keep the shared schema ID structural, and add any trigger-char discriminator at the HTML cache layer instead.

src/serializers/html/html.test.ts
```diff
@@ -379,20 +379,23 @@ describe('HTML Serializer', () => {
...
+                    '<p>Question: Who&#39;s the head of the Frontend team?</p><p>Answer: <span data-mention data-id="963827" data-label="Henning M">@Henning M</span></p>',
                 )
             })
 
             test('channel suggestion HTML output is correct', () => {
                 const htmlSerializer = createHTMLSerializer(
-                    getSchema([PlainTextKit, createSuggestionExtension('channel')]),
+                    getSchema([
+                        PlainTextKit,
+                        createSuggestionExtension('channel').configure({ triggerChar: '#' }),
```
doistbot, line 412: [P2] Switching this test to `#` verifies custom trigger output for one freshly created serializer, but it still doesn't exercise the cache-collision bug fixed in `computeSchemaId()`. Please add a nearby regression test that uses `getHTMLSerializerInstance()` with two schemas that differ only by `triggerChar` and proves they don't share the same instance/output.

doistbot (commented):
This follow-up PR thoroughly rounds out the suggestion HTML output by fixing plain-text serialization, schema caching, and centralizing URL scheme logic. These updates successfully close the loop on the previous implementation while notably improving the maintainability of the serialization pathways. A few remaining refinements have been noted regarding further deduplication of schema metadata traversal, properly escaping special trigger characters in string replacements, and isolating the cache discriminator to the HTML layer alongside an explicit regression test for cache collisions.

<sub>[Share Feedback](https://github.com/Doist/doistbot-app/issues/new?template=review-feedback.yml&title=Review+feedback+for+Doist%2Ftypist&pr_url=https%3A%2F%2Fgithub.com%2FDoist%2Ftypist%2Fpull%2F1342&datadog_url=https%3A%2F%2Fapp.datadoghq.com%2Flogs%3Fquery%3Dservice%253A%2528doistbot-app-sandbox%2520OR%2520doistbot-app%2529%2520%2540meta.deliveryId%253A181c2440-563c-11f1-8bff-622a5010a091%26agg_m%3Dcount%26agg_m_source%3Dbase%26agg_t%3Dcount%26clustering_pattern_field_path%3Dmessage%26cols%3Dhost%252Cservice%26fromUser%3Dtrue%26messageDisplay%3Dinline%26refresh_mode%3Dsliding%26storage%3Dhot%26stream_sort%3Ddesc%26viz%3Dstream%26live%3Dtrue) • [Review Logs](https://app.datadoghq.com/logs?query=service%3A%28doistbot-app-sandbox%20OR%20doistbot-app%29%20%40meta.deliveryId%3A181c2440-563c-11f1-8bff-622a5010a091&agg_m=count&agg_m_source=base&agg_t=count&clustering_pattern_field_path=message&cols=host%2Cservice&fromUser=true&messageDisplay=inline&refresh_mode=sliding&storage=hot&stream_sort=desc&viz=stream&live=true)</sub>

src/serializers/html/html.ts
rfgamaral: No longer applicable after the recent refactors. The trigger character is now read in a single place (`buildSuggestionSchemaInfo`). The other places that previously extracted it (`computeSchemaId` with its `''` fallback, and the plain-text `html.ts` path) either don't read it anymore (`computeSchemaId` is purely structural again) or consume it via `buildSuggestionSchemaInfo`'s `triggerCharByScheme` map. A dedicated helper isn't needed.

PR #1343, opened 2026-05-23
refactor(serializers)!: replace Turndown with unified for Markdown serializer

rfgamaral:
## Overview

Switches the underlying Markdown serializer from Turndown to the unified ecosystem. The public API is unchanged: `createMarkdownSerializer`, `getMarkdownSerializerInstance`, and `MarkdownSerializerReturnType` work exactly the same way as before.

### What consumers need to do

Update peer dependencies:

- **Remove:** `turndown`, `@types/turndown`
- **Add:** `rehype-parse`, `rehype-remark`, `remark-stringify`, `mdast-util-gfm-task-list-item`

### What consumers might notice in the output

The serialized Markdown is CommonMark-compliant. Most outputs match the previous Turndown output exactly, with two visible differences:

- Nested ordered lists are indented by 3 spaces (matches the list marker width) instead of 4. Nested unordered lists are indented by 2 spaces.
- Some edge cases in escape behavior may differ slightly (e.g., specific punctuation sequences).

These differences are cosmetic; the rendered HTML and behavior in any CommonMark renderer is unchanged.

## PR Checklist

## Test plan

This PR targets `next` for a pre-release before promoting to stable, so it can be smoke-tested both in Storybook and in production.

**In the deployed Storybook** (via the Netlify preview):

- Open the `Rich-text → Functions / Commands` story and use `insertMarkdownContent` to round-trip a comprehensive set of Markdown constructs.
- Or open any rich-text story, type freely, inspect the serialized Markdown output, then paste that output back as initial content to confirm the editor renders it the same way.

**As a pre-release in production:**

- Install the `@next` pre-release of `@doist/typist`.
- Update peer dependencies as listed above.
- Smoke test the editor and Markdown serialization end-to-end.


src/serializers/markdown/markdown.ts
```diff
@@ -1,17 +1,21 @@
...
-import { suggestion } from './plugins/suggestion'
-import { taskItem } from './plugins/task-item'
+import { rehypeImage } from './plugins/rehype-image'
+import { rehypeSuggestions } from './plugins/rehype-suggestions'
+import { rehypeTaskList } from './plugins/rehype-task-list'
+import { remarkStrikethrough } from './plugins/remark-strikethrough'
+import { remarkTaskList } from './plugins/remark-task-list'
 
 import type { Schema } from '@tiptap/pm/model'
+import type { Handle, Options as RehypeRemarkOptions } from 'hast-util-to-mdast'
```
doistbot: [P2] `hast-util-to-mdast` and `mdast-util-to-markdown` are not listed in `package.json`. Importing types directly from transitive (phantom) dependencies can cause type-checking errors for consumers using strict package managers like pnpm.

Consider importing `Options` from `rehype-remark` instead, and extracting the `Join` type from `RemarkStringifyOptions` (or letting TypeScript infer it):
```ts
import type { Options as RehypeRemarkOptions } from 'rehype-remark'
import type { Options as RemarkStringifyOptions } from 'remark-stringify'

type Join = NonNullable<RemarkStringifyOptions['join']>[number]
```

src/serializers/markdown/markdown.ts
```diff
@@ -40,52 +44,105 @@ type MarkdownSerializerInstanceById = {
...
+                    )
+                },
+            }),
+        },
+        join: [
+            // Force tight list rendering regardless of the `spread` attribute on lists or list
+            // items, so that simple list items are not separated by blank lines (matching the
+            // previous behavior and the common user expectation)
+            (_, __, parent) => {
+                if (parent.type === 'list' || parent.type === 'listItem') {
```
doistbot: [P1] Forcing `join` to `0` for every `listItem` child collapses multi-block list items. A valid item like `<li><p>First</p><p>Second</p></li>` will now serialize as a single paragraph continuation instead of preserving the paragraph break, so round-tripping loses structure. Limit this override to list-level joins, or keep the default spacing when a list item contains multiple block children.

src/serializers/markdown/markdown.ts
```diff
@@ -96,98 +153,60 @@ const INITIAL_TURNDOWN_OPTIONS: Turndown.Options = {
...
-            return markdownResult.replace(/ +$/gm, '')
+            // For plain-text editors, preserve runs of 2+ spaces by replacing them with
+            // non-breaking spaces before parsing (otherwise `rehype-parse` collapses them), and
+            // restoring them after serialization. This is required to preserve syntax like
+            // nested lists where leading indentation matters.
+            const input = isPlainText
+                ? html.replace(/ {2,}/g, (match) => match.replace(/ /g, '\u00a0'))
+                : html
+
+            const result = unifiedProcessor.processSync(input).toString().trimEnd()
```
doistbot, line 229: [P2] `trimEnd()` removes terminal blank lines from the serialized Markdown. In plain-text mode that drops trailing empty paragraphs/final line breaks that the `paragraph+` schema can represent, so copying the Markdown source no longer preserves EOF content. Keep the old per-line space cleanup, but don't trim terminal newlines.

src/serializers/markdown/markdown.ts
```diff
@@ -40,52 +44,105 @@ type MarkdownSerializerInstanceById = {
...
+        listItemIndent: 'one',
+        resourceLink: true,
+        handlers: {
+            // Output a plain newline for hard breaks instead of the CommonMark backslash syntax
+            // (i.e. `\\\n`), since `remark-breaks` treats plain newlines as hard breaks
+            break() {
+                return '\n'
+            },
+            // Return text values as-is without any Markdown escaping (plain-text only)
+            ...(isPlainText && {
```
doistbot: [P3] The dual conditional spread pattern can be simplified by defining `text(node)` once and checking `isPlainText` inside the function. This avoids duplicating the property assignment and simplifies readability.

src/serializers/markdown/plugins/remark-strikethrough.ts
```diff
@@ -0,0 +1,23 @@
...
+ * A remark plugin to add support for serializing the strikethrough extension from the GitHub
+ * Flavored Markdown (GFM) specification (i.e., `~~text~~`).
+ *
+ * This is a standalone plugin which makes use of the `mdast-util-gfm-strikethrough` package, and
+ * the implementation is inspired by the third-party `remark-gfm` plugin.
+ *
+ * The reason why we don't use `remark-gfm` directly is because we only want to opt into the
+ * specific GFM features the editor supports, rather than enabling all of them at once.
+ */
+function remarkStrikethrough(this: Processor) {
```
doistbot, line 15: [P2] This reimplements the same `toMarkdownExtensions` wiring that already exists in `src/serializers/html/plugins/remark-strikethrough.ts`. Reuse that plugin here, or move the shared strikethrough registration into a common module, so the HTML and Markdown serializers don't drift on future GFM changes.

src/serializers/markdown/markdown.ts
```diff
@@ -40,52 +44,105 @@ type MarkdownSerializerInstanceById = {
...
-        // @ts-ignore: The `Turndown.Node` type does not include `isBlock`
-        return node.isBlock ? '\n\n' : ''
-    },
+function getRehypeRemarkOptions(schema: Schema): RehypeRemarkOptions {
+    const handlers: Record<string, Handle> = {}
+
+    // When the schema does not support strikethrough, unwrap `<del>` elements into their
+    // children so the inner text is preserved without a `delete` mdast node being created
+    if (!schema.marks.strike) {
+        handlers.del = (state, node) => state.all(node)
```
doistbot, line 77: [P2] This hardcodes `del`, but the strike mark schema accepts `s`, `del`, and `strike` (the old Turndown path derived its tag set from `spec.parseDOM`). That breaks the schema-driven behavior here: unsupported `<s>`/`<strike>` input will no longer be unwrapped to plain text. Please derive these handlers from the mark's parse rules, or at least cover all supported strike tags.

src/serializers/markdown/markdown.ts
```diff
@@ -40,52 +44,105 @@ type MarkdownSerializerInstanceById = {
...
+            // (rich-text only)
+            ...(!isPlainText && {
+                text(node) {
+                    return (
+                        node.value
+                            // Escape backslashes that precede any punctuation mark, to prevent
+                            // the backslash itself from being interpreted as an escape sequence
+                            // for the subsequent character. It's important to apply this rule
+                            // first to avoid double escaping.
+                            .replace(new RegExp(`(\\\\${REGEX_PUNCTUATION.source})`, 'g'), '\\$1')
```
doistbot: [P2] Recompiling `new RegExp` on every invocation of the `text` handler adds unnecessary overhead. Move the regex instantiation outside the function (e.g., to module scope) to compile it only once.

src/serializers/markdown/plugins/rehype-suggestions.ts
```diff
@@ -0,0 +1,63 @@
...
+    return (...[tree]: Parameters<Transformer>): ReturnType<Transformer> => {
+        visit(tree, 'element', (node: HastNode) => {
+            if (!isHastElementNode(node, 'span')) {
+                return
+            }
+
+            // Find the data attribute matching a known suggestion URL scheme. Note that
+            // `rehype-parse` normalizes data attributes to camelCase in `properties`, so
+            // `data-mention` becomes `dataMention`.
+            const urlScheme = [...suggestionSchemaInfo.triggerCharByScheme.keys()].find(
```
doistbot: [P2] Spreading `.keys()` into a new array and invoking lodash's `camelCase` for every `<span>` node creates unnecessary allocation and string manipulation during the AST traversal. Precompute the camel-cased scheme names outside the `visit` loop.

src/serializers/markdown/plugins/rehype-image.ts
```diff
@@ -0,0 +1,49 @@
...
+ * `remark-stringify` handlers do not handle correctly:
+ *
+ * 1. Images with a Data URL source have their base64 content replaced with `NOT_SUPPORTED`,
+ *    to prevent enormous base64 strings from being dumped into the Markdown output.
+ * 2. Images with an empty or missing source are removed entirely, to prevent broken Markdown
+ *    syntax (i.e. `![]()` from being generated.
+ */
+function rehypeImage(): Transformer {
+    return (...[tree]: Parameters<Transformer>): ReturnType<Transformer> => {
+        // First, replace base64 content with `NOT_SUPPORTED` in Data URL images
```
doistbot: [P2] This plugin now walks the full HAST twice on every serialization for image-enabled schemas: once in `visit()` and again in `remove()`. The old Turndown rule handled images in a single pass, so this is avoidable extra work on a hot path. Please fold the sanitize/remove logic into one traversal over `img` nodes so image cleanup only scans the tree once.

src/serializers/markdown/markdown.test.ts
```diff
@@ -541,64 +541,64 @@ Strikethrough uses two tildes: ~~scratch this~~`,
...
 - And here's the third list item.`,
                 )
             })
 
             test('nested unordered lists Markdown output is correct', () => {
                 expect(markdownSerializer.serialize(HTML_NESTED_UNORDERED_LIST)).toBe(
                     `- Parent list element
-    - Sub-item 1
-    - Sub-item 2
+  - Sub-item 1
```
doistbot, line 582: [P2] These new 2-space nested list expectations are only covered on the rich-text path. The refactor moved plain-text whitespace preservation into a separate pre/post-processing step, but the plain-text suite still only checks a 4-space ordered-list indent. Add a plain-text nested-list case with the new 2/3-space indentation widths so regressions in that path don't slip through.

src/serializers/markdown/plugins/rehype-image.ts
```diff
@@ -0,0 +1,49 @@
...
+ * A rehype plugin to handle two edge cases for images that the default `rehype-remark` and
+ * `remark-stringify` handlers do not handle correctly:
+ *
+ * 1. Images with a Data URL source have their base64 content replaced with `NOT_SUPPORTED`,
+ *    to prevent enormous base64 strings from being dumped into the Markdown output.
+ * 2. Images with an empty or missing source are removed entirely, to prevent broken Markdown
+ *    syntax (i.e. `![]()` from being generated.
+ */
+function rehypeImage(): Transformer {
+    return (...[tree]: Parameters<Transformer>): ReturnType<Transformer> => {
```
doistbot: [P3] The explicit `ReturnType<Transformer>` annotation violates the TypeScript guideline to prefer inference when the compiler can easily deduce it. Additionally, the `(...[tree]: Parameters<Transformer>)` syntax is unnecessarily complex; consider simplifying this to `(tree: HastNode)`. (This also applies to the other `rehype-*` plugins.)

src/serializers/markdown/plugins/remark-strikethrough.ts
```diff
@@ -0,0 +1,23 @@
...
+    const data = this.data()
+
+    const toMarkdownExtensions = data.toMarkdownExtensions || (data.toMarkdownExtensions = [])
```
doistbot: [P3] The use of `||` for fallback assignment violates the TypeScript guideline to prefer `??` for nullish defaults. Please change to `??` here, as well as in `remark-task-list.ts`.

.oxlintrc.jsonc
```diff
@@ -49,7 +49,7 @@
         {
             // Unified plugins are invoked by the processor via `.use()`, which binds `this` to
             // the Processor instance; the bundler-`undefined`-`this` concern doesn't apply here
-            "files": ["src/serializers/html/plugins/*.ts"],
+            "files": ["src/serializers/html/plugins/*.ts", "src/serializers/markdown/plugins/*.ts"],
```
doistbot: [P2] This broadens the `no-this-in-exported-function` suppression to every markdown plugin, but only the remark plugins rely on unified binding `this`. The repo guidance is to fix or suppress issues at the narrowest scope possible; otherwise future rehype plugins in this folder will silently lose this check too. Please narrow this override to `remark-*.ts` (or the specific files) instead of the whole `plugins` directory.

doistbot (commented):
This PR successfully replaces Turndown with the unified ecosystem for Markdown serialization while keeping the public API intact. The migration to rehype and remark lays an excellent foundation for generating robust, CommonMark-compliant output. There are a few refinements to address before merging, primarily concerning data preservation with multi-block list items and trailing whitespace, optimizing AST traversals and regex compilations for performance, and tightening up type imports and linting suppressions.

<sub>[Share Feedback](https://github.com/Doist/doistbot-app/issues/new?template=review-feedback.yml&title=Review+feedback+for+Doist%2Ftypist&pr_url=https%3A%2F%2Fgithub.com%2FDoist%2Ftypist%2Fpull%2F1343&datadog_url=https%3A%2F%2Fapp.datadoghq.com%2Flogs%3Fquery%3Dservice%253A%2528doistbot-app-sandbox%2520OR%2520doistbot-app%2529%2520%2540meta.deliveryId%253Ab0a52460-56b8-11f1-98cb-c5c9653eee7d%26agg_m%3Dcount%26agg_m_source%3Dbase%26agg_t%3Dcount%26clustering_pattern_field_path%3Dmessage%26cols%3Dhost%252Cservice%26fromUser%3Dtrue%26messageDisplay%3Dinline%26refresh_mode%3Dsliding%26storage%3Dhot%26stream_sort%3Ddesc%26viz%3Dstream%26live%3Dtrue) • [Review Logs](https://app.datadoghq.com/logs?query=service%3A%28doistbot-app-sandbox%20OR%20doistbot-app%29%20%40meta.deliveryId%3Ab0a52460-56b8-11f1-98cb-c5c9653eee7d&agg_m=count&agg_m_source=base&agg_t=count&clustering_pattern_field_path=message&cols=host%2Cservice&fromUser=true&messageDisplay=inline&refresh_mode=sliding&storage=hot&stream_sort=desc&viz=stream&live=true)</sub>

src/serializers/markdown/markdown.ts
rfgamaral: The empty-paragraph concern doesn't hold: `hast-util-to-mdast` drops empty paragraphs upstream, so there's no trailing content to preserve. Probe confirms `<p>Foo</p><p></p>` and `<p>Foo</p><p></p><p></p>` both produce `Foo\n` regardless.

The only difference between `trimEnd()` and `/ +$/gm` is the single trailing `\n` `remark-stringify` adds for spec compliance. Keeping `trimEnd()` gives output parity with the previous Turndown-based serializer, which is the goal of this PR.

src/serializers/markdown/plugins/remark-strikethrough.ts
rfgamaral: The two plugins are intentionally different: the HTML serializer needs the full parser registration (`micromark` + `fromMarkdown` + `toMarkdown`) because it parses Markdown input, while the Markdown serializer only needs `toMarkdown` because it never parses Markdown. Reusing the HTML plugin here would mean the markdown serializer registers unused parser extensions for every instance it creates.

The drift risk is small in practice: both plugins pull from `mdast-util-gfm-strikethrough`, whose API is stable, and the extensions are one-line `push()` calls that don't accumulate complexity. The duplication is bounded and direction-specific, which I think is clearer than a shared module abstraction.

rfgamaral:
Not promoting this to stable. After integration testing in todoist-web (Doist/todoist-web#17894), two concerns surfaced:

1. ✅ **Peer dependency leakage.** Switching to the unified pipeline pushes its internal dependencies (`hast-util-*`, `mdast-util-*`, `rehype*`, `remark*`, `unified`, `unist-util-*`) onto consumers as peer dependencies. Most of these aren't imported anywhere else in consumer apps and shouldn't be a concern of theirs. They need to stay encapsulated inside `@doist/typist`.
    - Addressed in: https://github.com/Doist/typist/pull/1351
3. ❌ **Hot path performance.** Benchmarks show v11's unified-based serializer is 2-3× slower than v10's turndown-based one (~22 ms p95 on extreme content). In todoist-web, `getMarkdown()` runs every 100-150ms during typing for task descriptions and comments, so this adds material overhead.

More importantly, this migration doesn't make sense as an interim step given where Typist is headed. Tiptap v3 has built-in Markdown support that works directly on the ProseMirror doc - no HTML → hast → mdast roundtrip, no peer-dep surface to leak into consumers. Once we migrate to v3, the unified serializer goes away entirely, so all the work here, including the call-site optimizations to decouple the typing hot path, gets thrown out. Better to skip the intermediate hop and put the effort into v3 directly.

The `next` branch has been reverted to `main`. The `11.0.0-next.1` pre-release stays published as a reference point.

PR #1351, opened 2026-05-26
chore(deps)!: drop React 17 support and move runtime peer deps into dependencies

rfgamaral:
## Overview

Most of Typist's `peerDependencies` were packages it imports internally (`unified`, the `remark`/`rehype`/`mdast`/`micromark`/`unist` plugin chain, `turndown`, `linkifyjs`, `lodash-es`, etc.). They don't need a shared instance with the consumer, so they don't belong as peers.

Moves them all into `dependencies`, pinned to current resolved versions. `react` and `react-dom` stay as peers and narrow to `^18.0.0`. Drops the manual peer-deps install snippet from `README.md` and `CONTRIBUTING.md`; `npm install @doist/typist` is now the whole story.

**Breaking:** `react`/`react-dom` 17 are no longer supported. Consumers must be on `^18.0.0`.

## References

- Addresses the peer-dependency leakage concern raised in [#1343 (comment)](https://github.com/Doist/typist/pull/1343#issuecomment-4546328498).

## PR Checklist

- [x] Pull request title follows the [Conventional Commits Specification](https://www.conventionalcommits.org/)
- [x] Added/updated documentation to Storybook or `README.md`

## Test plan

Code review and CI are sufficient.

PR #1354, opened 2026-05-26
feat(deps): move shared application utilities back to peer dependencies

rfgamaral:
## Overview

Follow-up to #1351.

That PR moved every non-React peer dep into `dependencies`, which was correct for the pure-internal markdown chain (`unified`, `remark`/`rehype`/`mdast`/`micromark`/`unist` plugins) since consumers never import those directly, but wrong for `linkifyjs`, `emoji-regex`, and `lodash-es`. Those three are application-level utilities consumers commonly use in their own code, so leaving them as regular deps causes silent bundle duplication when consumer pins diverge from Typist's pins. Confirmed in practice on [Doist/todoist-web#17906](https://github.com/Doist/todoist-web/pull/17906): two copies of `linkifyjs` ended up in the task editor chunk.

Moves those three back to `peerDependencies` with broad ranges so npm enforces alignment at install time and bundlers ship a single copy. The pure internals stay where they are.

## PR Checklist

## Test plan

Code review and CI are sufficient.

### DIFF
--- a/package.json
+++ b/package.json
@@ -154,7 +154,7 @@
     },
     "peerDependencies": {
         "emoji-regex": "^10.2.1",
-        "linkifyjs": "^4.1.1",
+        "linkifyjs": "^4.3.2",
         "lodash-es": "^4.17.21",
         "react": "^18.0.0",
         "react-dom": "^18.0.0"
--- a/package.json
+++ b/package.json
@@ -87,9 +87,6 @@
         "@tiptap/pm": "2.27.2",
         "@tiptap/react": "2.27.2",
         "@tiptap/suggestion": "2.27.2",
-        "emoji-regex": "10.6.0",
-        "linkifyjs": "4.3.3",
-        "lodash-es": "4.17.23",
         "mdast-util-gfm-autolink-literal": "2.0.1",
         "mdast-util-gfm-strikethrough": "2.0.0",
         "micromark-extension-gfm-autolink-literal": "2.1.0",
@@ -156,6 +153,9 @@
         "vitest": "4.1.7"
     },
     "peerDependencies": {
+        "emoji-regex": "^10.2.1",
+        "linkifyjs": "^4.1.1",
+        "lodash-es": "^4.17.21",
         "react": "^18.0.0",
         "react-dom": "^18.0.0"
     }
--- a/package-lock.json
+++ b/package-lock.json
@@ -111,7 +111,7 @@
             },
             "peerDependencies": {
                 "emoji-regex": "^10.2.1",
-                "linkifyjs": "^4.1.1",
+                "linkifyjs": "^4.3.2",
                 "lodash-es": "^4.17.21",
                 "react": "^18.0.0",
                 "react-dom": "^18.0.0"
--- a/package-lock.json
+++ b/package-lock.json
@@ -40,9 +40,6 @@
                 "@tiptap/pm": "2.27.2",
                 "@tiptap/react": "2.27.2",
                 "@tiptap/suggestion": "2.27.2",
-                "emoji-regex": "10.6.0",
-                "linkifyjs": "4.3.3",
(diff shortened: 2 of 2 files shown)


doistbot (commented):
Thanks @rfgamaral for your contribution 😇. Good catch on the bundle duplication issue with the shared application utilities.

Few things worth tightening:
* Bump the `linkifyjs` peer dependency minimum to `^4.3.2` to match `@tiptap/extension-link` and avoid pulling in a duplicate copy anyway.
* Reconsider making these package-wide peer dependencies without subpath exports, as the root barrel re-exports will force all consumers to install them even if they only use `TypistEditor`.

<sub>[Share Feedback](https://github.com/Doist/doistbot-app/issues/new?template=review-feedback.yml&title=Review+feedback+for+Doist%2Ftypist&pr_url=https%3A%2F%2Fgithub.com%2FDoist%2Ftypist%2Fpull%2F1354&datadog_url=https%3A%2F%2Fapp.datadoghq.com%2Flogs%3Fquery%3Dservice%253A%2528doistbot-app-sandbox%2520OR%2520doistbot-app%2529%2520%2540meta.deliveryId%253Ab09aca80-5931-11f1-8d50-a2aa89ba6c70%26agg_m%3Dcount%26agg_m_source%3Dbase%26agg_t%3Dcount%26clustering_pattern_field_path%3Dmessage%26cols%3Dhost%252Cservice%26fromUser%3Dtrue%26messageDisplay%3Dinline%26refresh_mode%3Dsliding%26storage%3Dhot%26stream_sort%3Ddesc%26viz%3Dstream%26live%3Dtrue) • [Review Logs](https://app.datadoghq.com/logs?query=service%3A%28doistbot-app-sandbox%20OR%20doistbot-app%29%20%40meta.deliveryId%3Ab09aca80-5931-11f1-8d50-a2aa89ba6c70&agg_m=count&agg_m_source=base&agg_t=count&clustering_pattern_field_path=message&cols=host%2Cservice&fromUser=true&messageDisplay=inline&refresh_mode=sliding&storage=hot&stream_sort=desc&viz=stream&live=true)</sub>

package.json
```diff
@@ -156,6 +153,9 @@
         "vitest": "4.1.7"
     },
     "peerDependencies": {
+        "emoji-regex": "^10.2.1",
+        "linkifyjs": "^4.1.1",
```
doistbot: <a href="#">P2</a>  `^4.1.1` is looser than the runtime requirement already in `@tiptap/extension-link` (`linkifyjs@^4.3.2`). In apps that satisfy this peer with `4.1.x` or `4.2.x`, npm/pnpm will still install a second `linkifyjs` copy for Tiptap, so the bundle-duplication problem comes back. Please raise this peer minimum to at least `^4.3.2` (or whatever minimum all internal runtime users require).

package.json
```diff
@@ -156,6 +153,9 @@
         "vitest": "4.1.7"
     },
     "peerDependencies": {
+        "emoji-regex": "^10.2.1",
```
doistbot, line 156: <a href="#">P2</a>  Making these packages peers leaks them through the whole package, not just the APIs that use them. `src/index.ts` statically re-exports modules like `createSuggestionExtension`, `createHTMLSerializer`, `extend-word-range`, and the rich-text extensions, and those modules import `lodash-es`, `linkifyjs`, or `emoji-regex`. In the published unbundled ESM build, any consumer importing `@doist/typist` has to resolve those re-exported modules during linking, so even apps that only use `TypistEditor` now need to declare all three peers. Either keep them as regular dependencies, or split the affected APIs into separate subpath exports before moving them to peers.

rfgamaral: Technically correct, but Typist is Doist-focused per the README, and our consumers don't actually run into this. Subpath exports are too big a refactor for a problem nobody has. Leaving as-is.

frankieyan (commented):
✅

PR #1372, opened 2026-06-11
fix: replace wrapping empty textblock when inserting block-only Markdown content

rfgamaral:
## Overview

The `insertMarkdownContentAt` command (which also powers `insertMarkdownContent` and the `PasteMarkdown` extension) inserts parsed content with `tr.replaceRange`, without the empty-textblock handling that the official `insertContentAt` command implements. When block-only content is inserted at a cursor position inside an empty paragraph, ProseMirror's fitting algorithm cannot always merge the content into the paragraph, and the empty paragraph lingers next to the inserted content.

This PR mirrors the official [`insertContentAt` behaviour](https://github.com/ueberdosis/tiptap/blob/v2.27.1/packages/core/src/commands/insertContentAt.ts#L130-L143): when only block content is inserted at a cursor position within an empty (non-code) textblock, the insertion range is expanded to replace the wrapping textblock itself.

Note that all built-in nodes already replace the empty paragraph correctly: their parsed slices are either closed (images, horizontal rules) or open through layers that ProseMirror's fitting algorithm can close and join cleanly against a paragraph boundary (code blocks, blockquotes, lists, which bottom out in textblocks). As such, this change has no observable effect with the current built-in schema. The bug becomes observable with nodes whose slices are open through multiple non-textblock structural layers — concretely, the upcoming table support (`table > tableRow > tableHeader > paragraph`, where pasting a Markdown table left an empty paragraph next to the table), and potentially custom extensions passed to `<TypistEditor>` with similar nesting.

## PR Checklist

## Test plan

Regression-tested the existing behaviour (which should remain unchanged) on the preview Storybook:

- Paste Markdown content of each kind into the **empty** editor (plain text, `- list`, `> quote`, ` ```code``` `, `image`, `---`)
    - [ ] Observe that the content is inserted without a lingering empty paragraph, and the Markdown output matches the pasted input
- Type some text, then paste block Markdown content after it
    - [ ] Observe that the content is inserted below the text, with no empty paragraph in between

### DIFF
--- a/src/extensions/core/extra-editor-commands/commands/insert-markdown-content-at.ts
+++ b/src/extensions/core/extra-editor-commands/commands/insert-markdown-content-at.ts
@@ -74,9 +74,10 @@ function insertMarkdownContentAt(
                 options.parseOptions,
             )
 
-            let isOnlyBlockContent = true
+            // Check if the parsed content is non-empty and composed of block nodes only (the
+            // initial value guards against an empty insert deleting the wrapping textblock below)
+            let isOnlyBlockContent = content.content.childCount > 0
 
-            // Check if the parsed content is composed of block nodes only
             content.content.forEach((node) => {
                 isOnlyBlockContent = isOnlyBlockContent ? node.isBlock : false
             })
--- a/src/extensions/core/extra-editor-commands/commands/insert-markdown-content-at.ts
+++ b/src/extensions/core/extra-editor-commands/commands/insert-markdown-content-at.ts
@@ -62,7 +62,7 @@ function insertMarkdownContentAt(
             }
 
             // Get the start and end positions from the provided position
-            const { from, to } =
+            let { from, to } =
                 typeof position === 'number'
                     ? { from: position, to: position }
                     : { from: position.from, to: position.to }
@@ -73,6 +73,29 @@ function insertMarkdownContentAt(
                 parseHtmlToElement(htmlContent),
                 options.parseOptions,
             )
+
+            let isOnlyBlockContent = true
+
+            // Check if the parsed content is composed of block nodes only
+            content.content.forEach((node) => {
+                isOnlyBlockContent = isOnlyBlockContent ? node.isBlock : false
+            })
+
+            // Expand the insertion range to replace the wrapping empty textblock when only block
+            // content is inserted at a cursor position, mirroring the official `insertContentAt`
+            // command behaviour (e.g., a table pasted into an empty paragraph should replace the
+            // paragraph, instead of keeping the empty paragraph below the table)
+            if (from === to && isOnlyBlockContent) {
+                const { parent } = tr.doc.resolve(from)
+
+                const isEmptyTextBlock =
+                    parent.isTextblock && !parent.type.spec.code && !parent.childCount
+
+                if (isEmptyTextBlock) {
+                    from -= 1
+                    to += 1
+                }
+            }
 
             // Inserts the content into the editor while preserving the current selection
             tr.replaceRange(from, to, content)


src/extensions/core/extra-editor-commands/commands/insert-markdown-content-at.ts
```diff
@@ -74,6 +74,29 @@ function insertMarkdownContentAt(
                 options.parseOptions,
             )
 
+            let isOnlyBlockContent = true
```
doistbot: <big><kbd>🟡 P2</kbd></big>  If `markdown` parses to an empty fragment, `isOnlyBlockContent` will default to `true` (since the `forEach` loop won't run). This causes an empty insert to unintentionally delete the wrapping empty paragraph instead of doing nothing.

Initializing this against the fragment's child count ensures an empty insert safely remains a no-op:
```ts
let isOnlyBlockContent = content.content.childCount > 0
```

src/extensions/core/extra-editor-commands/commands/insert-markdown-content-at.ts
```diff
@@ -74,6 +74,29 @@ function insertMarkdownContentAt(
...
+            // Check if the parsed content is composed of block nodes only
+            content.content.forEach((node) => {
+                isOnlyBlockContent = isOnlyBlockContent ? node.isBlock : false
+            })
+
+            // Expand the insertion range to replace the wrapping empty textblock when only block
+            // content is inserted at a cursor position, mirroring the official `insertContentAt`
+            // command behaviour (e.g., a table pasted into an empty paragraph should replace the
+            // paragraph, instead of keeping the empty paragraph below the table)
+            if (from === to && isOnlyBlockContent) {
```
doistbot, line 89: <big><kbd>🟡 P2</kbd></big>  This new branch is the reason for the fix, but the PR only lists Storybook smoke tests for built-in markdown nodes, which the description says don't exercise it. Please add an automated regression test for the real failure mode: with a custom nested block schema (same shape as the upcoming table support), call `insertMarkdownContentAt` from an empty paragraph and assert the wrapper paragraph is replaced instead of left adjacent to the inserted block.

doistbot (commented):
This cleanly updates `insertMarkdownContentAt` to replace overlapping empty textblocks when inserting block-only content, nicely mirroring Tiptap's standard behavior.

Few things worth tightening:
- Guard against empty parsed fragments by checking the child count, ensuring an empty string insertion doesn't unintentionally delete the current empty paragraph.
- Add an automated regression test using a nested block schema (like the upcoming tables) to explicitly cover this new path, since built-in nodes won't trigger it.

<sub>[Share Feedback](https://github.com/Doist/doistbot-app/issues/new?template=review-feedback.yml&title=Review+feedback+for+Doist%2Ftypist&pr_url=https%3A%2F%2Fgithub.com%2FDoist%2Ftypist%2Fpull%2F1372&datadog_url=https%3A%2F%2Fapp.datadoghq.com%2Flogs%3Fquery%3Dservice%253A%2528doistbot-app-sandbox%2520OR%2520doistbot-app%2529%2520%2540meta.deliveryId%253A9b2c9b30-65af-11f1-8414-e600cf2f5989%26agg_m%3Dcount%26agg_m_source%3Dbase%26agg_t%3Dcount%26clustering_pattern_field_path%3Dmessage%26cols%3Dhost%252Cservice%26fromUser%3Dtrue%26messageDisplay%3Dinline%26refresh_mode%3Dsliding%26storage%3Dhot%26stream_sort%3Ddesc%26viz%3Dstream%26live%3Dtrue) • [Review Logs](https://app.datadoghq.com/logs?query=service%3A%28doistbot-app-sandbox%20OR%20doistbot-app%29%20%40meta.deliveryId%3A9b2c9b30-65af-11f1-8414-e600cf2f5989&agg_m=count&agg_m_source=base&agg_t=count&clustering_pattern_field_path=message&cols=host%2Cservice&fromUser=true&messageDisplay=inline&refresh_mode=sliding&storage=hot&stream_sort=desc&viz=stream&live=true)</sub>

src/extensions/core/extra-editor-commands/commands/insert-markdown-content-at.ts
rfgamaral: Not adding tests at this moment.

src/extensions/core/extra-editor-commands/commands/insert-markdown-content-at.ts
rfgamaral: Good catch — an empty insert would have hit ProseMirror's empty-slice shortcut (`deleteRange`) and deleted the wrapping paragraph. Addressed in 13d3ad9 with the suggested initialization.

PR #1373, opened 2026-06-11
feat!: add GFM table support to the rich-text editor

rfgamaral:
## Overview

This PR adds table support to the rich-text editor, scoped to what the GitHub Flavored Markdown (GFM) table syntax can represent, mirroring the approach Tiptap v3 takes with its official Markdown support (same stock extensions and node names, GFM-only serialization), which should ease a future migration.

### Editing

- The `RichTextKit` now registers the official `Table`/`TableRow`/`TableHeader`/`TableCell` extensions by default, with a `table` option to customize or disable (`table: false`)
- Table cells are restricted to a single paragraph of content, since the GFM table syntax cannot represent multiple blocks within a cell, making sure the editor cannot produce tables that lose content when serialized to Markdown (block toggles don't apply within cells, and `editor.can()` reports that correctly for menu/toolbar disabling)
- All resizing-related options are hidden from `RichTextTableOptions`, since column widths cannot be represented in the GFM table syntax
- Tables are not available in singleline documents, since tables always serialize to multiple lines of Markdown (note that tables also require a `paragraph` node in the editor schema, as table cells are restricted to paragraph content)
- A new input rule inserts a table by typing two or more consecutive pipe characters followed by a space (the pipes delimit the number of columns, e.g. `||| ` inserts a two-column table with a heading row and one body row); new rows can be created by pressing <kbd>Tab</kbd> at the last cell (built-in behaviour)

### Serialization

- Markdown → HTML: a new `remark-table` plugin adds support for the GFM table syntax (following the same pattern as the existing GFM strikethrough/autolink plugins); column alignment is parsed but not preserved, as Tiptap cells don't support alignment
- HTML → Markdown: a new Turndown `table` plugin outputs GFM pipe tables, with pipe characters escaped, an empty heading row added to tables without one (the GFM syntax requires it), and hard breaks within cells serialized as literal `<br>` elements (which GFM renders as line breaks)
- A new `rehype-table` plugin restores raw `<br>` elements within table cells (at any depth, e.g. within styled text) into actual hard breaks, making multi-line cells round-trip losslessly (raw HTML remains escaped text everywhere else); the hard break handling on both sides is only enabled when the schema supports hard breaks
- Text that looks like a `<br>` element is escaped in the Markdown output, so it isn't confused with the literal `<br>` elements emitted for hard breaks within table cells

### Pasting

- Pasted Markdown tables and HTML tables (e.g., from spreadsheet apps) are now inserted as native tables in rich-text editors, and as such the `PasteHTMLTableAsString` extension is no longer registered by the `RichTextKit` (it remains in the `PlainTextKit`, where tables cannot exist)
- Tables pasted without a heading row — both HTML tables from spreadsheet apps (composed of data cells only) and heading-less Markdown pipe rows (which the GFM syntax wouldn't parse as a table at all) — get an empty heading row added at paste time, keeping what is seen in the editor consistent with the serialized Markdown

### Storybook

- New "Insert Table" toolbar button which prompts for the number of columns/rows (invalid input falls back to 3), disabled when the editor doesn't support tables or when the cursor is inside a table (tables cannot be nested)
- Base table styling, plus a gap cursor positioning fix (with the default styles, the gap cursor after a trailing image or table renders at the previous block's top-left corner instead of at the gap)
- New "Gap Cursor" and "Tables" sections in the Tips & Tricks → Styling documentation

## ⚠️ Breaking Changes

- Tables are **enabled by default**: Markdown content with GFM pipe-table syntax now renders as native tables in any `RichTextKit` editor without a code change on the consumer's side (disable with `table: false`)
- The `pasteHTMLTableAsString` option was **removed** from `RichTextKitOptions` (the option remains available on `PlainTextKitOptions`)

## PR Checklist

## Test plan

- Open the preview Storybook deployed to Netlify, and the `Rich-text → Default` story
- Click the "Insert Table" toolbar button, and accept the default prompts
    - [x] Observe that a 3×3 table with a heading row is inserted
- With the caret inside the table, check the "Insert Table" toolbar button
    - [x] Observe that the button is disabled (tables cannot be nested)
- Type content into cells, navigating with <kbd>Tab</kbd>/<kbd>Shift+Tab</kbd>, and press <kbd>Tab</kbd> at the last cell
    - [x] Observe that a new row is created, and the Markdown output is a GFM pipe table
- In an empty editor, type `||| ` (pipes followed by a space)
    - [x] Observe that a two-column table with a heading row and one body row is inserted
- Paste a GFM table (e.g. the Markdown output from above) into an empty editor
    - [x] Observe that a native table is inserted, with no lingering empty paragraph
- Copy cells from a spreadsheet app (e.g., Google Sheets) and paste them into the editor
    - [x] Observe that a native table is inserted, with an empty heading row added on top of the pasted rows
- Paste a heading-less Markdown table (pipe rows without the `| --- |` delimiter row, e.g. `| One | Two |`) into the editor
    - [ ] Observe that a native table is inserted, with an empty heading row added on top of the pasted rows
- With the caret in the last cell, press <kbd>↓</kbd>, then type some text
    - [x] Observe that the cursor renders below the table, and the text is typed below the table
- Open the `Rich-text → Singleline` story and paste a GFM table
    - [x] Observe that the table Markdown is inserted as literal text (tables are not available in singleline documents)


src/serializers/html/plugins/rehype-table.ts
```diff
@@ -0,0 +1,40 @@
...
+function rehypeTable(): Transformer {
+    return (...[tree]: Parameters<Transformer>): ReturnType<Transformer> => {
+        visit(tree, 'element', (node: HastNode) => {
+            if (!isHastElementNode(node, 'th') && !isHastElementNode(node, 'td')) {
+                return
+            }
+
+            // Replace raw `<br>` nodes (which would otherwise be output as escaped text) with
+            // actual hard break elements
+            node.children = node.children.map((child) =>
```
doistbot: <big><kbd>🟠 P1</kbd></big>  `node.children.map` only checks the direct children of the `th` or `td` element. If a hard break is enclosed in formatting tags (e.g., `| **hello<br>world** |`), the `raw` node for `<br>` will be nested inside an element like `em` or `strong` in the HAST. Because this isn't performing a nested/recursive sub-tree search, the `<br>` will be missed and subsequently serialized as escaped text (e.g., `&lt;br&gt;`) rather than resolving to an actual hard break. Consider walking the specific nodes using another `visit` call like `visit(node, 'raw', ...)` to properly hit nested `<br>` raw elements.

doistbot (commented):
This introduces GFM-compliant table support to the editor, carefully mirroring Tiptap's approach to keep Markdown serialization predictable.

Few things worth tightening:
- **Hard breaks:** A few edge cases around `<br>` serialization need addressing, such as finding nested line breaks in formatted text, escaping literal `<br>` inputs, respecting schemas without `hardBreak`s, and narrowing the AST traversal to protect performance.
- **Schema configurations:** The schema builder throws if `paragraph` is disabled while tables are enabled, and the singleline exclusion behavior needs test coverage.
- **Contextual UI:** The table insert button remains enabled inside existing tables, so it needs wiring to `editor.can().insertTable()` to disable dynamically.

<details>
<summary>Optional follow-up notes (2)</summary>

- **[P3] src/serializers/markdown/plugins/table.ts:80**: Consider mapping these tag arrays to uppercase once during initialization (e.g., `...parseDOM).map(tag => tag.toUpperCase())`). Turndown invokes the `filter` callbacks for every node in the document; hoisting `.toUpperCase()` out of those callbacks and `isHeadingRow` avoids repetitive string allocations and allows for a simpler `.includes(node.nodeName)` on the serialization hot path.
- **[P3] src/serializers/markdown/markdown.test.ts:961**: Consider adding a test verifying what the Markdown serializer outputs when given HTML tables while the `table` extension is disabled. The adjacent HTML serializer test suite specifically checks this fallback behavior (`test('tables Markdown syntax is ignored when the table extension is disabled')`), making it worth covering the equivalent Markdown serialization path here.

</details>

<sub>[Share Feedback](https://github.com/Doist/doistbot-app/issues/new?template=review-feedback.yml&title=Review+feedback+for+Doist%2Ftypist&pr_url=https%3A%2F%2Fgithub.com%2FDoist%2Ftypist%2Fpull%2F1373&datadog_url=https%3A%2F%2Fapp.datadoghq.com%2Flogs%3Fquery%3Dservice%253A%2528doistbot-app-sandbox%2520OR%2520doistbot-app%2529%2520%2540meta.deliveryId%253A70e831e0-65b4-11f1-8717-8f1b0d614a53%26agg_m%3Dcount%26agg_m_source%3Dbase%26agg_t%3Dcount%26clustering_pattern_field_path%3Dmessage%26cols%3Dhost%252Cservice%26fromUser%3Dtrue%26messageDisplay%3Dinline%26refresh_mode%3Dsliding%26storage%3Dhot%26stream_sort%3Ddesc%26viz%3Dstream%26live%3Dtrue) • [Review Logs](https://app.datadoghq.com/logs?query=service%3A%28doistbot-app-sandbox%20OR%20doistbot-app%29%20%40meta.deliveryId%3A70e831e0-65b4-11f1-8717-8f1b0d614a53&agg_m=count&agg_m_source=base&agg_t=count&clustering_pattern_field_path=message&cols=host%2Cservice&fromUser=true&messageDisplay=inline&refresh_mode=sliding&storage=hot&stream_sort=desc&viz=stream&live=true)</sub>

src/extensions/rich-text/rich-text-kit.ts
```diff
@@ -321,6 +324,25 @@ const RichTextKit = Extension.create<RichTextKitOptions>({
...
+
+        if (this.options.table !== false && !isSinglelineDocument) {
+            extensions.push(
+                RichTextTable.configure(this.options?.table),
+                TableRow,
+
+                // Restrict cells to a single paragraph (instead of one or more blocks), since the
+                // GFM table syntax cannot represent multiple blocks within a cell, ensuring the
+                // editor cannot produce tables that lose content when serialized to Markdown
+                TableHeader.extend({ content: 'paragraph' }),
```
doistbot, line 342: <big><kbd>🟠 P1</kbd></big>  This makes `paragraph` a hidden requirement of `RichTextKit`, but the option is still independently disableable. `RichTextKit.configure({ paragraph: false })` now builds a schema where table cells require a missing `paragraph` node, which throws during schema creation once tables stay enabled by default. Please either force `table: false` when `paragraph` is disabled, or keep `paragraph` required while tables are enabled.

src/serializers/markdown/plugins/table.ts
```diff
@@ -0,0 +1,174 @@
...
+        // cannot represent multiple lines within a cell, while raw `<br>` elements are rendered
+        // as line breaks by GFM, and parsed back into hard breaks by the HTML serializer
+        const cellSelector = [...headerTags, ...cellTags].join(', ')
+
+        turndown.addRule('tableCellHardBreak', {
+            filter(node) {
+                return node.nodeName === 'BR' && node.closest(cellSelector) !== null
+            },
+            replacement() {
+                return '<br>'
```
doistbot: <big><kbd>🟡 P2</kbd></big>  Emitting a bare `<br>` here makes hard breaks indistinguishable from a cell whose literal text is `<br>`. That content now round-trips back as an actual line break because `rehypeTable` restores every raw `<br>` inside cells. Please use a representation that can't collide with user text, or escape literal `<br>` text before emitting the Markdown table.

stories/typist-editor/decorators/typist-editor-decorator/typist-editor-toolbar.tsx
```diff
@@ -221,6 +252,12 @@ function TypistEditorToolbar({ editor }: TypistEditorToolbarProps) {
             withDividerBefore: true,
             onClick: handleImageButtonClick,
         },
+        {
+            'aria-label': 'Insert Table',
+            disabled: !editor.schema.nodes.table,
```
doistbot: <big><kbd>🟡 P2</kbd></big>  The `disabled` check only verifies if the table extension is registered. This means the button remains enabled even when the user's cursor is in a location where a table cannot be inserted (e.g., inside an existing table). To match the context-aware disabling of other buttons (like `insertImage`), consider using Tiptap's dynamic command evaluation instead: `disabled: !editor.can().insertTable()` or `disabled: !editor.can().insertTable({ rows: 1, cols: 1 })`.

src/serializers/markdown/markdown.ts
```diff
@@ -150,6 +151,23 @@ function createMarkdownSerializer(schema: Schema): MarkdownSerializerReturnType
...
     }
 
+    // Add rules for `table` if the corresponding nodes exists in the schema
+    if (
+        schema.nodes.table &&
+        schema.nodes.tableRow &&
+        schema.nodes.tableHeader &&
+        schema.nodes.tableCell
+    ) {
+        turndown.use(
```
doistbot, line 166: <big><kbd>🟡 P2</kbd></big>  The new table plugin is wired in whenever the table nodes exist, but its `tableCellHardBreak` rule is unconditional. That puts the serializer pair out of sync for `RichTextKit.configure({ hardBreak: false })`: the Markdown serializer can emit literal `<br>` inside cells, while the HTML serializer for the same schema won't restore them because `rehypeTable` is gated on `schema.nodes.hardBreak`. Please pass hard-break support into the table plugin and skip that rule when the schema doesn't include `hardBreak`.

src/serializers/html/plugins/rehype-table.ts
```diff
@@ -0,0 +1,40 @@
...
+ *
+ * Since the GFM table syntax cannot represent multiple lines within a cell, the Markdown
+ * serializer outputs hard breaks inside table cells as literal `<br>` elements (which GFM renders as line
+ * breaks). However, the HTML serializer persists raw HTML as escaped text (i.e. raw HTML is not
+ * interpreted), and thus this plugin restores raw `<br>` elements found within table cells into
+ * actual hard breaks.
+ */
+function rehypeTable(): Transformer {
+    return (...[tree]: Parameters<Transformer>): ReturnType<Transformer> => {
+        visit(tree, 'element', (node: HastNode) => {
```
doistbot, line 19: <big><kbd>🟡 P2</kbd></big>  This adds a full extra HAST walk to every rich-text Markdown→HTML serialization now that tables are enabled by default. `TypistEditor` calls `htmlSerializer.serialize(content)` whenever the controlled `content` prop changes, so even documents with no tables pay for `visit(tree, 'element', ...)`, and each cell also gets a new `children` array via `.map(...)`. If you visit only `raw` nodes (or otherwise narrow the traversal to table descendants) and mutate only matching `<br>` entries, you can keep the common path from traversing the whole tree.

src/extensions/rich-text/rich-text-kit.ts
```diff
@@ -321,6 +324,25 @@ const RichTextKit = Extension.create<RichTextKitOptions>({
             extensions.push(RichTextStrikethrough.configure(this.options?.strike))
         }
 
+        // Tables are not supported in singleline documents because, although a table is a single
+        // block node (which would technically fit the singleline schema), it serializes to
+        // multiple lines of Markdown, breaking the singleline contract
+        const isSinglelineDocument =
+            this.options.document !== false && this.options.document?.multiline === false
+
+        if (this.options.table !== false && !isSinglelineDocument) {
```
doistbot: <big><kbd>🟡 P2</kbd></big>  The logic to disable the `table` extension in singleline documents is a crucial protection against breaking the singleline contract, but it lacks test coverage. Consider adding a test (for example, by testing `computeSchemaId` in `src/helpers/schema.test.ts`) to verify that the table node extensions are successfully omitted from the schema when `RichTextKit.configure({ document: { multiline: false } })` is used.

src/serializers/html/plugins/rehype-table.ts
rfgamaral: Addressed the actionable half in acad261 together with the nested-`<br>` fix: the plugin now mutates only matching raw nodes in place (no per-cell `children` array allocations), within a single traversal. The full-tree walk itself is inherent to any rehype transform — `visit` traverses all nodes regardless of the type filter, and the pipeline already performs several full walks (e.g. `rehype-minify-whitespace`, `rehype-stringify`) — so narrowing beyond this isn't meaningful.

stories/typist-editor/decorators/typist-editor-decorator/typist-editor-toolbar.tsx
rfgamaral: Addressed with `isActive(editorState, 'table')` (plus the existing schema check) rather than the suggested `editor.can().insertTable()`, which can't work here: `insertTable` returns `true` unconditionally (all placement work is inside `if (dispatch)`), so the dry-run never reports failure — and on schemas without the table extension, `can().insertTable` is `undefined` and calling it throws.

rfgamaral:
_Sorry `Frontend Product` for the mass ping on this one, adding the group with additional reviewers at the same time didn't work as I expected it would._

gnapse (approved):
You had me at "This PR adds table support…" and I did not want to read any further.

a cartoon character is holding a bunch of money and the words shut up and take are above him

Everything from the test plan checks out, except one point that I left unchecked. Not sure if I was doing something wrong. But it's also a feature that I do not think if it's that important to support, even if it's true that it does not work, which I'm not sure of.

Anyway, like I said. This is great!

src/extensions/rich-text/rich-text-kit.ts
rfgamaral: Heads-up: this guard was removed again in ccbab54. Reviewing the equivalent change for list extensions in #1374 surfaced that keying registration off the kit's local `paragraph` option breaks the canonical replace-pattern (`[RichTextKit.configure({ paragraph: false }), CustomParagraph]`), since the kit cannot know whether a `paragraph` node will exist in the final schema. The agreed contract: kit options govern what the kit registers, and schema validity is enforced at schema creation, where ProseMirror's error already names the missing node — a truly paragraph-less rich-text editor is not a supported configuration. See #1374 (closed) for the full reasoning.

rfgamaral:
> Everything from the test plan checks out, except one point that I left unchecked. Not sure if I was doing something wrong.

@gnapse You probably copied the example from the test plan, and pasted into the editor with `Cmd+V`, correct? That will paste as a code mark. You need to paste-as-text with `Cmd+Shift+V`, and that will work (just tested it myself).

amix:
Skimmed the code, and this looks lovely, @rfgamaral. Very excited about this! 🔥

PR #1374, opened 2026-06-11
fix: do not register list extensions when the paragraph extension is disabled

rfgamaral:
## Overview

Configuring `RichTextKit` with `paragraph: false` crashes during schema creation:

```
SyntaxError: No node type or group 'paragraph' found (in content expression 'paragraph block*')
```

The root cause is that the kit still registers the list extensions when the `Paragraph` extension is disabled, and `ListItem` declares `content: 'paragraph block*'`, which references the now-missing `paragraph` node type. The bulleted/ordered list nodes would also dangle without `listItem`, and `ListKeymap` is pointless without list nodes.

This PR fixes the crash by silently skipping the registration of all list-related extensions (`BulletList`, `ListItem`, `ListKeymap`, and `OrderedList`) when the `paragraph` option is disabled, and documents this behavior in the `RichTextKitOptions` doc comments. The default schema (i.e. with paragraphs enabled) is unchanged.

This bug is pre-existing (it has been around since the kit was introduced, but nobody disables paragraphs), and it's unrelated to other in-flight PRs.

## PR Checklist

- [ ] Added/updated documentation to Storybook or `README.md`

## Test plan

- `getSchema([RichTextKit.configure({ paragraph: false })])` used to throw the `SyntaxError` above; it now builds a valid schema without the `bulletList`, `orderedList`, and `listItem` nodes (covered by a new unit test in `src/helpers/schema.test.ts`)
- `getSchema([RichTextKit])` still produces the exact same schema as before (covered by the existing `#computeSchemaId` unit test)
- Full test suite passes: 156 tests across 8 files

### DIFF
--- a/src/extensions/rich-text/rich-text-kit.ts
+++ b/src/extensions/rich-text/rich-text-kit.ts
@@ -67,7 +67,8 @@ type RichTextKitOptions = {
     bold: Partial<BoldOptions> | false
 
     /**
-     * Set options for the `BulletList` extension, or `false` to disable.
+     * Set options for the `BulletList` extension, or `false` to disable. This extension is always
+     * disabled when the `paragraph` option is disabled.
      */
     bulletList: Partial<RichTextBulletListOptions> | false
 
@@ -132,17 +133,20 @@ type RichTextKitOptions = {
     link: Partial<RichTextLinkOptions> | false
 
     /**
-     * Set options for the `ListItem` extension, or `false` to disable.
+     * Set options for the `ListItem` extension, or `false` to disable. This extension is always
+     * disabled when the `paragraph` option is disabled.
      */
     listItem: Partial<ListItemOptions> | false
 
     /**
-     * Set options for the `ListKeymap` extension, or `false` to disable.
+     * Set options for the `ListKeymap` extension, or `false` to disable. This extension is always
+     * disabled when the `paragraph` option is disabled.
      */
     listKeymap: Partial<ListKeymapOptions> | false
 
     /**
-     * Set options for the `OrderedList` extension, or `false` to disable.
+     * Set options for the `OrderedList` extension, or `false` to disable. This extension is always
+     * disabled when the `paragraph` option is disabled.
      */
     orderedList: Partial<RichTextOrderedListOptions> | false
 
@@ -209,7 +213,10 @@ const RichTextKit = Extension.create<RichTextKitOptions>({
             extensions.push(Bold.configure(this.options?.bold))
         }
 
-        if (this.options.bulletList !== false) {
+        // List items are restricted to paragraph-first content (i.e. `'paragraph block*'`), so a
+        // schema with list nodes but no `paragraph` node would throw during schema creation, which
+        // is why all list extensions are skipped when the `Paragraph` extension is disabled
+        if (this.options.bulletList !== false && this.options.paragraph !== false) {
             extensions.push(RichTextBulletList.configure(this.options?.bulletList))
         }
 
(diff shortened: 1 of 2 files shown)


src/extensions/rich-text/rich-text-kit.ts
```diff
@@ -209,7 +213,10 @@ const RichTextKit = Extension.create<RichTextKitOptions>({
             extensions.push(Bold.configure(this.options?.bold))
         }
 
-        if (this.options.bulletList !== false) {
+        // List items are restricted to paragraph-first content (i.e. `'paragraph block*'`), so a
+        // schema with list nodes but no `paragraph` node would throw during schema creation, which
+        // is why all list extensions are skipped when the `Paragraph` extension is disabled
+        if (this.options.bulletList !== false && this.options.paragraph !== false) {
```
doistbot, line 219: <big><kbd>🟡 P2</kbd></big>  This makes `RichTextKit.configure({ paragraph: false })` look like a supported default config, but the default Markdown path still breaks with that schema. `CopyMarkdownSource` is still registered, and `getMarkdownSerializerInstance()` unconditionally builds the paragraph rule from `schema.nodes.paragraph`, so `Mod-Shift-C` (or any direct `createMarkdownSerializer(getSchema(...))` call) will throw once it hits a paragraph-less schema. Either keep rejecting paragraph-less rich-text schemas, or gate the Markdown/copy extensions the same way.

doistbot, line 219: <big><kbd>🟡 P2</kbd></big>  This couples list registration to `RichTextKit`'s own `paragraph` option instead of the final extension set. A composition like `getSchema([RichTextKit.configure({ paragraph: false }), Paragraph])` used to be valid because the schema still had a `paragraph` node, but this guard now drops all list support anyway. That’s hard to recover from through Typist’s public API, because `src/index.ts` exports `RichTextKit` but not the internal `RichTextBulletList` / `RichTextOrderedList` wrappers. Please keep list registration possible when callers supply their own `paragraph` extension, rather than keying it off the local option flag.

doistbot (commented):
This fixes the schema creation crash when paragraphs are disabled in `RichTextKit` by skipping the dependent list extensions.

Few things worth tightening:
- Gate the Markdown serialization and copy extensions similarly, as they still unconditionally depend on `schema.nodes.paragraph` and will crash without it.
- Preserve list registration when callers supply their own custom `Paragraph` extension, rather than completely dropping list support based on the local kit option.

<sub>[Share Feedback](https://github.com/Doist/doistbot-app/issues/new?template=review-feedback.yml&title=Review+feedback+for+Doist%2Ftypist&pr_url=https%3A%2F%2Fgithub.com%2FDoist%2Ftypist%2Fpull%2F1374&datadog_url=https%3A%2F%2Fapp.datadoghq.com%2Flogs%3Fquery%3Dservice%253A%2528doistbot-app-sandbox%2520OR%2520doistbot-app%2529%2520%2540meta.deliveryId%253A98303f90-65b9-11f1-8fbb-19f7b72e1a07%26agg_m%3Dcount%26agg_m_source%3Dbase%26agg_t%3Dcount%26clustering_pattern_field_path%3Dmessage%26cols%3Dhost%252Cservice%26fromUser%3Dtrue%26messageDisplay%3Dinline%26refresh_mode%3Dsliding%26storage%3Dhot%26stream_sort%3Ddesc%26viz%3Dstream%26live%3Dtrue) • [Review Logs](https://app.datadoghq.com/logs?query=service%3A%28doistbot-app-sandbox%20OR%20doistbot-app%29%20%40meta.deliveryId%3A98303f90-65b9-11f1-8fbb-19f7b72e1a07&agg_m=count&agg_m_source=base&agg_t=count&clustering_pattern_field_path=message&cols=host%2Cservice&fromUser=true&messageDisplay=inline&refresh_mode=sliding&storage=hot&stream_sort=desc&viz=stream&live=true)</sub>

rfgamaral:
Closing after analyzing the review comments together: they showed the approach itself is wrong, not the details. `paragraph: false` exists primarily to *replace* the kit's `Paragraph` extension with a custom one (the canonical kit pattern), and keying list registration off the kit's local option silently breaks that composition — the kit cannot know whether a `paragraph` node will exist in the final schema. The right contract is: kit options govern what the kit registers, and schema validity is enforced at schema creation — where ProseMirror's error already names the missing node. A truly paragraph-less rich-text editor is not a configuration we support (or that anyone uses), so failing fast there is correct, and no code change is needed.

PR #1384, opened 2026-06-17
fix(serializers): preserve inline Markdown characters in suggestion labels

rfgamaral:
## Overview

Suggestion chips built from a Markdown link (like `[Title](hashtag://id)`) lost part of their label whenever the title contained inline-Markdown characters. A title such as ``The future of `code` `` produced a chip labelled `#The future of ` (everything from the backtick onward was dropped), and a title that *started* with such characters produced no chip at all and was left as a plain link. The same happened with emphasis (`*`, `_`), strikethrough (`~~`), and autolinks. The cause is that the label is parsed as inline Markdown, which splits it into several pieces, and only the first piece was being read.

This reads the full text of the link when building the chip, so the label stays intact no matter how the Markdown parser splits it, and a chip is now created even when its label starts with inline Markdown. It also escapes those characters when a chip is serialized back to Markdown, so a label keeps its exact text when content is parsed again, for example while editing existing content. Escaping applies to rich-text editors only, since plain-text editors keep the label verbatim when parsing it back.

## Reference

- Comms thread: https://comms.todoist.com/69/ch/CaVwVqsJ5M73bmSTT4geh/t/Cc8abjjHYirbpCMDSjTkW/

## PR Checklist

## Test plan

> [!TIP]
> Open the preview Storybook deployed to Netlify and select the `Typist Editor / Rich-text → Default` story.

- Place the cursor in the editor and paste the block below as plain text with `Cmd/Ctrl+Shift+V`

````
[The future of `code`](hashtag://1)

[Use *bold* please](hashtag://1)

[Status ~~done~~ now](hashtag://1)

[Ping <a@b.com> today](hashtag://1)

[`code` first](hashtag://1)
````

- [ ] Observe that the editor shows five chips with their full labels, matching:

```
#The future of code

#Use bold please

#Status done now

#Ping a@b.com today

#code first
```

## Demo

**Before / After**

https://github.com/user-attachments/assets/e2300016-4f8a-4d7a-bae3-8789ee82e11b

### DIFF
--- a/src/serializers/markdown/markdown.test.ts
+++ b/src/serializers/markdown/markdown.test.ts
@@ -416,6 +416,18 @@ Answer: [Doist Frontend](channel://190200)`)
                     'Hey [Alice](mention://123) and [Bob](mention://456), check [General](channel://789)',
                 )
             })
+
+            test('inline Markdown characters in a suggestion label are not escaped', () => {
+                const markdownSerializer = createMarkdownSerializer(
+                    getSchema([PlainTextKit, createSuggestionExtension('channel')]),
+                )
+
+                expect(
+                    markdownSerializer.serialize(
+                        '<p><span data-channel data-id="1" data-label="The future of `code`">#The future of `code`</span></p>',
+                    ),
+                ).toBe('[The future of `code`](channel://1)')
+            })
         })
     })
 
@@ -1100,6 +1112,30 @@ Answer: [Doist Frontend](channel://190200)`)
                 ).toBe(
                     'Hey [Alice](mention://123) and [Bob](mention://456), check [General](channel://789)',
                 )
+            })
+
+            test('inline Markdown characters in a suggestion label are escaped', () => {
+                const markdownSerializer = createMarkdownSerializer(
+                    getSchema([RichTextKit, createSuggestionExtension('channel')]),
+                )
+
+                expect(
+                    markdownSerializer.serialize(
+                        '<p><span data-channel data-id="1" data-label="Use *bold* and _more_">#Use *bold* and _more_</span></p>',
+                    ),
+                ).toBe('[Use \\*bold\\* and \\_more\\_](channel://1)')
+
+                expect(
+                    markdownSerializer.serialize(
+                        '<p><span data-channel data-id="1" data-label="~~struck~~ label">#~~struck~~ label</span></p>',
+                    ),
+                ).toBe('[\\~\\~struck\\~\\~ label](channel://1)')
+
+                expect(
+                    markdownSerializer.serialize(
+                        '<p><span data-channel data-id="1" data-label="a[b]c &lt;x&gt;">#a[b]c &lt;x&gt;</span></p>',
+                    ),
+                ).toBe('[a\\[b\\]c \\<x>](channel://1)')
             })
         })
     })
--- a/src/serializers/markdown/plugins/suggestion.ts
+++ b/src/serializers/markdown/plugins/suggestion.ts
@@ -1,4 +1,4 @@
-import { getSuggestionUrlScheme } from '../../../helpers/serializer'
+import { escapeSuggestionLabel, getSuggestionUrlScheme } from '../../../helpers/serializer'
 
 import type { NodeType } from '@tiptap/pm/model'
 import type Turndown from 'turndown'
@@ -8,8 +8,9 @@ import type Turndown from 'turndown'
  * extension factory function.
  *
  * @param nodeType The node object that matches this rule.
+ * @param isPlainText Specifies if the schema represents a plain-text document.
  */
-function suggestion(nodeType: NodeType): Turndown.Plugin {
+function suggestion(nodeType: NodeType, isPlainText: boolean): Turndown.Plugin {
     const attributeType = getSuggestionUrlScheme(nodeType)
 
     return (turndown: Turndown) => {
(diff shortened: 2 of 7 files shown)

### CODE REVIEW

File: src/serializers/markdown/plugins/suggestion.ts [hunk 1]
```diff
@@ -21,7 +22,13 @@ function suggestion(nodeType: NodeType): Turndown.Plugin {
                 const label = String((node as Element).getAttribute('data-label'))
                 const id = String((node as Element).getAttribute('data-id'))
 
-                return `[${label}](${attributeType}://${id})`
+                // Rich-text editors parse the label back as inline Markdown, so its Markdown
+                // characters are escaped to keep the label intact across a serialize then parse
+                // round-trip. Plain-text editors keep the label verbatim when parsing it back, so
+                // it must not be escaped.
+                const serializedLabel = isPlainText ? label : escapeSuggestionLabel(label)
```
doistbot, line 29: <big><kbd>🟡 P2</kbd></big>  Escaping the label only in rich-text mode makes the emitted Markdown incompatible with the plain-text loader. For example, this now serializes `a[b]c` as `[a\[b\]c](channel://1)` and `back\slash` as `[back\\slash](channel://1)`, but `createHTMLSerializerForPlainTextEditor()` doesn't unescape or even recognize those forms, so loading that same Markdown into a plain-text editor shows a plain link or extra backslashes instead of a chip. Please keep the serialized format readable by both modes, or teach the plain-text parser to handle these escapes before changing the output.

src/serializers/markdown/markdown.test.ts
```diff
@@ -1101,6 +1113,30 @@ Answer: [Doist Frontend](channel://190200)`)
...
+                    markdownSerializer.serialize(
+                        '<p><span data-channel data-id="1" data-label="~~struck~~ label">#~~struck~~ label</span></p>',
+                    ),
+                ).toBe('[\\~\\~struck\\~\\~ label](channel://1)')
+
+                expect(
+                    markdownSerializer.serialize(
+                        '<p><span data-channel data-id="1" data-label="a[b]c &lt;x&gt;">#a[b]c &lt;x&gt;</span></p>',
+                    ),
+                ).toBe('[a\\[b\\]c \\<x>](channel://1)')
```
doistbot, line 1138: <big><kbd>🟡 P2</kbd></big>  The `escapeSuggestionLabel` regex escapes 8 characters, but this test only exercises 6 (`*`, `_`, `~`, `[`, `]`, `<`). The backslash and backtick are untested in the serialization direction — and the backslash has zero coverage in either test file, despite the function's JSDoc specifically calling it out ("The backslash is matched first so it is never misread as an escape sequence"). A label like `a\*b` depends on the backslash being escaped to round-trip correctly; without it, the serialized `a\*b` re-parses as `a*b` (backslash consumed as a Markdown escape). Adding a case like `data-label="back\\slash"` → `'[back\\\\slash](channel://1)'` to this existing test would close the gap.

doistbot (commented):
This PR fixes suggestion chips built from Markdown links losing label text when the title contains inline-Markdown characters, by reading the full link text and escaping those characters during serialization in rich-text mode.

Few things worth tightening:
- Escaping labels only in rich-text mode means the emitted Markdown (e.g. `[a\[b\]c](channel://1)`) isn't understood by the plain-text loader, which doesn't unescape these forms—either keep the format readable by both modes or teach the plain-text parser to handle the escapes.
- Add test coverage for backslash escaping in `escapeSuggestionLabel`, both in the serialization direction and round-trip scenarios—currently the backslash has no unit test coverage despite being called out as a special concern in the JSDoc.
- Consider checking for a valid `triggerChar` before calling `getHastTextContent(node)` so the AST traversal only runs for known suggestion links rather than every `<a>` element.

<details>

- **[P3] src/serializers/html/plugins/rehype-suggestions.ts:38**: `getHastTextContent(node)` traverses the entire AST of the link, and is currently called for all `<a>` elements (including regular web links). To avoid this unnecessary overhead, consider checking for a valid `triggerChar` first and only computing `label` if the link is a known suggestion.

</details>

<sub>[Share Feedback](https://github.com/Doist/doistbot-app/issues/new?template=review-feedback.yml&title=Review+feedback+for+Doist%2Ftypist&pr_url=https%3A%2F%2Fgithub.com%2FDoist%2Ftypist%2Fpull%2F1384&datadog_url=https%3A%2F%2Fapp.datadoghq.com%2Flogs%3Fquery%3Dservice%253A%2528doistbot-app-sandbox%2520OR%2520doistbot-app%2529%2520%2540meta.deliveryId%253Ad1050cc0-6a90-11f1-8576-a2cbb83b8854%26agg_m%3Dcount%26agg_m_source%3Dbase%26agg_t%3Dcount%26clustering_pattern_field_path%3Dmessage%26cols%3Dhost%252Cservice%26fromUser%3Dtrue%26messageDisplay%3Dinline%26refresh_mode%3Dsliding%26storage%3Dhot%26stream_sort%3Ddesc%26viz%3Dstream%26live%3Dtrue) • [Review Logs](https://app.datadoghq.com/logs?query=service%3A%28doistbot-app-sandbox%20OR%20doistbot-app%29%20%40meta.deliveryId%3Ad1050cc0-6a90-11f1-8576-a2cbb83b8854&agg_m=count&agg_m_source=base&agg_t=count&clustering_pattern_field_path=message&cols=host%2Cservice&fromUser=true&messageDisplay=inline&refresh_mode=sliding&storage=hot&stream_sort=desc&viz=stream&live=true)</sub>

src/serializers/markdown/plugins/suggestion.ts
rfgamaral: Plain-text and rich-text editors don't load the same Markdown content, so this cross-mode case doesn't occur in practice. Within a single editor type the format round-trips correctly, and rich-text needs the escaping to do so, so keeping it rich-text-only is intentional.

src/serializers/markdown/markdown.test.ts
rfgamaral: Added backtick and backslash cases to the rich-text escaping test, so all eight escaped characters are now covered in the serialization direction.

PR #1386, opened 2026-06-18
fix(rich-text): Degrade pasted HTML tables to string when tables unavailable

engfragui:
## Overview

Follows up to https://github.com/Doist/typist/pull/1373.

Register the `PasteHTMLTableAsString` extension in `RichTextKit` when native table support is unavailable (tables disabled or singleline document), so pasted HTML tables from spreadsheets or GFM become space-separated paragraphs instead of running together.

## PR Checklist

- [ ] Added/updated documentation to Storybook or `README.md`

## Test plan

- Visit https://deploy-preview-1386--doist-typist.netlify.app/
- Open Rich-text > Singleline
- Paste a row from a spreadsheet ([sample spreadsheet](https://docs.google.com/spreadsheets/d/1iU_DkSdrHf0wxeb2nHSStLXUrHsGhDTDzQnRLIk4d9E/edit?usp=sharing))
- [ ] Verify that the space between cells is preserved

## Demo

| Screenshot 2026-06-18 at 17 20 32 https://typist.doist.dev/?path=/story/typist-editor-rich-text--singleline | Screenshot 2026-06-18 at 17 20 24 https://deploy-preview-1386--doist-typist.netlify.app/?path=/story/typist-editor-rich-text--singleline |

### DIFF
--- a/src/extensions/rich-text/rich-text-kit.ts
+++ b/src/extensions/rich-text/rich-text-kit.ts
@@ -19,6 +19,7 @@ import { Typography } from '@tiptap/extension-typography'
 
 import { BLOCKQUOTE_EXTENSION_PRIORITY } from '../../constants/extension-priorities'
 import { CopyMarkdownSource } from '../shared/copy-markdown-source'
+import { PasteHTMLTableAsString } from '../shared/paste-html-table-as-string'
 import { PasteSinglelineText } from '../shared/paste-singleline-text'
 
 import { BoldAndItalics } from './bold-and-italics'
@@ -159,6 +160,16 @@ type RichTextKitOptions = {
      * Set to `false` to disable the `PasteEmojis` extension.
      */
     pasteEmojis: false
+
+    /**
+     * Set to `false` to disable the `PasteHTMLTableAsString` extension, which converts pasted HTML
+     * tables (e.g. from spreadsheets or GitHub Flavored Markdown) into space-separated paragraphs.
+     *
+     * This extension is only registered when native table support is unavailable (i.e. `table` is
+     * `false`, or the document is singleline). When tables are enabled, pasted tables become native
+     * tables instead, so this option has no effect.
+     */
+    pasteHTMLTableAsString: false
 
     /**
      * Set to `false` to disable the `PasteMarkdown` extension.
@@ -342,6 +353,12 @@ const RichTextKit = Extension.create<RichTextKitOptions>({
                 TableHeader.extend({ content: 'paragraph' }),
                 TableCell.extend({ content: 'paragraph' }),
             )
+        } else if (this.options.pasteHTMLTableAsString !== false) {
+            // When native tables aren't available (disabled, or a singleline document), gracefully
+            // degrade pasted HTML tables (from spreadsheets or GFM) into space-separated paragraphs
+            // instead of letting the cell contents run together. This mirrors the `PlainTextKit`,
+            // which always registers this extension since it never supports native tables
+            extensions.push(PasteHTMLTableAsString)
         }
 
         if (this.options.text !== false) {
--- a/src/extensions/rich-text/rich-text-kit.test.ts
+++ b/src/extensions/rich-text/rich-text-kit.test.ts
@@ -1,0 +1,56 @@
+import { RichTextKit } from './rich-text-kit'
+
+import type { RichTextKitOptions } from './rich-text-kit'
+
+/**
+ * Resolves the extensions that `RichTextKit` registers for a given set of options and returns their
+ * names. Invokes `addExtensions()` directly (rather than instantiating an `Editor`) so the registration
+ * logic can be asserted without building a ProseMirror view.
+ */
+function getRegisteredExtensionNames(options?: Partial<RichTextKitOptions>): string[] {
+    const kit = options ? RichTextKit.configure(options) : RichTextKit
+    const childExtensions =
+        kit.config.addExtensions?.call({
+            name: kit.name,
+            options: kit.options,
+            storage: kit.storage,
+            parent: undefined,
+        }) ?? []
+
+    return childExtensions.map((extension) => extension.name)
+}
+
+describe('Extension: RichTextKit', () => {
+    describe('PasteHTMLTableAsString registration', () => {
+        test('registers native tables (not the paste-as-string fallback) by default', () => {
+            const names = getRegisteredExtensionNames()
+
+            expect(names).toContain('table')
+            expect(names).not.toContain('pasteHTMLTableAsString')
+        })
+
+        test('registers the paste-as-string fallback when tables are disabled', () => {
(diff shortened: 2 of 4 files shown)


rfgamaral:
@engfragui The only thing I'd probably change here is to make this a `fix` instead of `feat`, it's effectively **fixing** broken behavior, not **adding** a new feature. Other than that, it looks great to me 👏

rfgamaral (approved):
Looks good to me 🚀 Great job 👍

src/extensions/rich-text/rich-text-kit.ts
```diff
@@ -342,6 +353,12 @@ const RichTextKit = Extension.create<RichTextKitOptions>({
                 TableHeader.extend({ content: 'paragraph' }),
                 TableCell.extend({ content: 'paragraph' }),
             )
+        } else if (this.options.pasteHTMLTableAsString !== false) {
```
doistbot, line 356: <big><kbd>🟡 P2</kbd></big>  Auto-registering `PasteHTMLTableAsString` whenever `table: false` breaks consumers that disable the built-in table extension so they can provide their own. In a setup like `RichTextKit.configure({ table: false })` plus another table extension, this high-priority transform rewrites pasted spreadsheet/GFM HTML into paragraphs before the replacement extension can parse it, so pasted tables degrade unexpectedly. Please avoid treating `table: false` as "no table support" by default here, or gate this fallback behind an explicit opt-in for that case.

src/extensions/rich-text/rich-text-kit.ts
```diff
@@ -342,6 +353,12 @@ const RichTextKit = Extension.create<RichTextKitOptions>({
...
+            // When native tables aren't available (disabled, or a singleline document), gracefully
+            // degrade pasted HTML tables (from spreadsheets or GFM) into space-separated paragraphs
+            // instead of letting the cell contents run together. This mirrors the `PlainTextKit`,
+            // which always registers this extension since it never supports native tables
+            extensions.push(PasteHTMLTableAsString)
```
doistbot, line 361: <big><kbd>🟡 P2</kbd></big>  In singleline editors and `table: false` editors, this now installs `PasteHTMLTableAsString` for every HTML paste. That plugin's `transformPastedHTML` path always runs `DOMParser.parseFromString(...)` before it knows whether the fragment even contains a `<table>`, so large non-table HTML pastes now pay a full parse in these editors too. Please add a cheap `/<table[\s>]/i` guard in the shared transformer before expanding this registration.

doistbot (commented):
Registers the `PasteHTMLTableAsString` extension in `RichTextKit` when native table support is unavailable (tables disabled or singleline document), so pasted HTML tables from spreadsheets or GFM degrade to space-separated paragraphs instead of running together.

Few things worth tightening:
- Auto-registering `PasteHTMLTableAsString` whenever `table: false` can break consumers that disable the built-in table extension to provide their own—the fallback transform will rewrite pasted HTML before the custom extension can parse it. Consider treating `table: false` differently from "no table support," or gate the fallback behind an explicit opt-in for that case.
- The extension's `transformPastedHTML` runs a full `DOMParser.parseFromString` on every HTML paste, even when the fragment contains no `<table>`. Adding a cheap `/\u003ctable[\s\u003e]/i` guard before the parse would avoid the overhead for large non-table pastes in singleline and table-disabled editors.

<sub>[Share Feedback](https://github.com/Doist/doistbot-app/issues/new?template=review-feedback.yml&title=Review+feedback+for+Doist%2Ftypist&pr_url=https%3A%2F%2Fgithub.com%2FDoist%2Ftypist%2Fpull%2F1386&datadog_url=https%3A%2F%2Fapp.datadoghq.com%2Flogs%3Fquery%3Dservice%253A%2528doistbot-app-sandbox%2520OR%2520doistbot-app%2529%2520%2540meta.deliveryId%253A0157e7e0-6b2b-11f1-8b28-81b1c8035021%26agg_m%3Dcount%26agg_m_source%3Dbase%26agg_t%3Dcount%26clustering_pattern_field_path%3Dmessage%26cols%3Dhost%252Cservice%26fromUser%3Dtrue%26messageDisplay%3Dinline%26refresh_mode%3Dsliding%26storage%3Dhot%26stream_sort%3Ddesc%26viz%3Dstream%26live%3Dtrue) • [Review Logs](https://app.datadoghq.com/logs?query=service%3A%28doistbot-app-sandbox%20OR%20doistbot-app%29%20%40meta.deliveryId%3A0157e7e0-6b2b-11f1-8b28-81b1c8035021&agg_m=count&agg_m_source=base&agg_t=count&clustering_pattern_field_path=message&cols=host%2Cservice&fromUser=true&messageDisplay=inline&refresh_mode=sliding&storage=hot&stream_sort=desc&viz=stream&live=true)</sub>

src/extensions/rich-text/rich-text-kit.ts
rfgamaral: @doistbot This PR restores previous behavior before proper table support was added, so your breaking assessment is not necessarily true. Besides, consumers that want to provide their own table extension can set `RichTextKit.configure({ table: false, pasteHTMLTableAsString: false })`, and do so. I don't think this is worth addressing (cc @engfragui).

src/extensions/rich-text/rich-text-kit.ts
rfgamaral: @engfragui This one might be worth addressing, should be a simple check inside the `PasteHTMLTableAsString` extension, I think.

engfragui: @rfgamaral Addressed in 738693470199f705e564af338e9277d8bbefcb49 before reading your comment... but didn't add a test for it (adding now)

src/extensions/rich-text/rich-text-kit.ts
doistbot: Fair point. I checked the history and you're right — before v14.0.0 added native table support, `PasteHTMLTableAsString` was always registered by `RichTextKit`, so this restores that prior default rather than introducing something new. The narrow window where a consumer set `table: false` in v14.0.0+ (when the fallback was unregistered) to use a custom table extension is small, and they'd now just be back to the pre-v14 paste behavior.

The opt-out path is also solid: `pasteHTMLTableAsString` is a documented option on `RichTextKitOptions`, the test at `rich-text-kit.test.ts:46` explicitly covers the `{ table: false, pasteHTMLTableAsString: false }` case, and the JSDoc on the option explains when the extension registers and how to disable it. A consumer wiring in a custom table extension has a clear, documented escape hatch.

I'll drop the P2. The one thing worth keeping in mind going forward: the `transformPastedHTML` hook runs at priority `1005`, ahead of `PasteMarkdown` (`1001`) and the default extension priority. So any future custom table extension that also relies on `transformPastedHTML` would need consumers to remember the opt-out — but that's an edge case and the documentation covers it. LGTM as-is.

Issue #849, opened 2024-07-22
Switch custom useEditor hook for the official one provided by Tiptap


thales-fukuda:
With the release of Tiptap `v2.5`,  the [`useEditor` hook now includes an option to render the editor on the first render](https://github.com/ueberdosis/tiptap/pull/5161). This makes the custom hook used here obsolete, which was included to solve the same issue and is currently outdated compared to the one provided by Tiptap.

Switching back to the official hook and providing an option to opt out could also fix #532 and #737.

Issue #737, opened 2024-04-14
Does not work with Next.js


18601673727:
```tsx
"use client"

import { PlainTextKit } from "@doist/typist"

const TypistEditor = dynamic(() => import("@doist/typist").then((mod) => mod.TypistEditor), {
  ssr: false,
})

// ...Later in render()

<TypistEditor placeholder="A full rich-text editor, be creative…" content={value} extensions={[PlainTextKit]} />
```

result in error:
```
Unhandled Runtime Error
TypeError: this.docView is null
```

mwolf1989:
Same for me.

Created a use client TypistEditor Component, try to use it in a page but get this error:

```
TypeError: Cannot read properties of null (reading 'updateOuterDeco')

Call Stack
EditorView.updateStateInner
node_modules/prosemirror-view/dist/index.js (5305:1)
EditorView.update
node_modules/prosemirror-view/dist/index.js (5234:1)
EditorView.setProps
node_modules/prosemirror-view/dist/index.js (5248:1)
Editor.createNodeViews
node_modules/@tiptap/core/dist/index.js (3894:1)
PureEditorContent.init
node_modules/@tiptap/react/dist/index.js (52:1)
PureEditorContent.componentDidMount
node_modules/@tiptap/react/dist/index.js (35:1)
```

rfgamaral:
Closed by https://github.com/Doist/typist/pull/1401.

Issue #532, opened 2023-11-20
Error:  Cannot read properties of null (reading 'updateOuterDeco')


thales-fukuda:
# Bug description

While running the [example](https://github.com/Doist/typist#usage) presented on the README for this project in dev mode, the code threw the error below. 

I tried to use the `use-editor` hook provided by Tiptap instead of the custom one for this project and the code ran well. I think the problem may be related to the fact that while hot-reloading, Tiptap tries to access a destroyed editor.

The `use-editor` hook was written more than a year ago, and since then Tiptap had multiple updates. Specifically,  https://github.com/ueberdosis/tiptap/pull/4000 might be related to it.

# Error:
```
Cannot read properties of null (reading 'updateOuterDeco')
TypeError: Cannot read properties of null (reading 'updateOuterDeco')
    at EditorView.updateStateInner (http://localhost:3000/static/js/bundle.js:99248:24)
    at EditorView.update (http://localhost:3000/static/js/bundle.js:99181:10)
    at EditorView.setProps (http://localhost:3000/static/js/bundle.js:99193:10)
    at Editor.createNodeViews (http://localhost:3000/static/js/bundle.js:57999:15)
    at PureEditorContent.init (http://localhost:3000/static/js/bundle.js:62223:14)
    at PureEditorContent.componentDidMount (http://localhost:3000/static/js/bundle.js:62204:10)
    at invokeLayoutEffectMountInDEV (http://localhost:3000/static/js/bundle.js:36371:26)
    at invokeEffectsInDev (http://localhost:3000/static/js/bundle.js:38289:15)
    at commitDoubleInvokeEffectsInDEV (http://localhost:3000/static/js/bundle.js:38270:9)
    at flushPassiveEffectsImpl (http://localhost:3000/static/js/bundle.js:38045:9)
```

# Version information:

Package version: 2.1.3
React: 18.2.0

PR #1387, opened 2026-06-19
refactor!: migrate to Tiptap's built-in useEditor hook

rfgamaral:
## Overview

Typist's custom `useEditor` hook exists for one reason only: to stop Tiptap's hook from returning `null` on the first render. Tiptap solved that itself in v2.5 with the `immediatelyRender` option, and its built-in hook has improved a lot since, so there's no reason to keep maintaining our outdated fork. This PR removes the custom hook and switches `TypistEditor` to Tiptap's built-in `useEditor`.

With the built-in hook the editor is created once and kept for the component's lifetime instead of being recreated whenever props change. That makes `content`, `extensions`, and `placeholder` initialization-only: they're read when the editor first mounts, and runtime updates go through the editor instance instead, for example `editor.commands.setContent()` for content, or reading changing data reactively from inside an extension. In exchange, `editable` updates in place, event handler props always use their latest value so they no longer need to be memoized, and the editor no longer re-renders on every change.

Dropping the outdated hook also clears the `null` editor crashes it caused under StrictMode, hot reload, and SSR. This is a breaking change for consumers that updated the initialization-only props at runtime; the migration steps are in the commit's `BREAKING CHANGE` notes and the new Component Lifecycle and Reactive Editor State documentation.

## Reference

- Completes the migration tracked in #849
- Addresses the `null` editor crashes reported in #532 and #737

## PR Checklist

## Test plan

> [!TIP]
> Open the preview Storybook deployed to Netlify. For the event handler checks, keep the `Actions` panel open.

### Editor smoke test

- Type, edit, and format text, paste some content, and trigger a suggestion by typing `@`
    - [x] Observe that typing, formatting, pasting, and suggestions all behave as before
- Open the `Plain-text → Default` story and type a few lines
    - [x] Observe that the plain-text editor behaves as before

### Event handlers and lifecycle

- Open the `Rich-text → Event Handlers` story
- Type a few characters in the editor
    - [x] Observe that `onUpdate`, `onTransaction`, and `onSelectionUpdate` fire in the `Actions` panel
    - [x] Observe that `onKeyDown` fires for each keystroke
- Click inside the editor content
    - [x] Observe that `onClick` fires
- Click `Unmount editor`
    - [x] Observe that `onDestroy` fires and the editor is removed while the toolbar stays in place
- Click `Mount editor`
    - [x] Observe that the editor reappears and accepts input again


src/components/typist-editor.tsx
```diff
@@ -321,62 +305,74 @@ const TypistEditor = forwardRef<TypistEditorRef, TypistEditorProps>(function Typ
...
+        autofocus: autoFocus ? 'end' : false,
+        content: htmlContent,
+        editable,
+        editorProps: {
+            attributes: {
+                'data-typist-editor': 'true',
+                ...(isPlainTextDocument(schema)
+                    ? { 'data-plain-text': 'true' }
+                    : { 'data-rich-text': 'true' }),
+                'aria-readonly': String(!editable),
```
doistbot: <big><kbd>🟡 P2</kbd></big>  `aria-readonly` and other ARIA attributes passed here will become out of sync if their corresponding props (like `editable` or `ariaLabel`) change at runtime.

Since `useEditor` only captures `editorProps` on the initial mount, toggling `editable` updates `contenteditable` (via the `syncEditableState` effect below) but leaves `aria-readonly` stale on the DOM element.

Consider adding an effect to sync these attributes dynamically when they change, for example by using `editor.setOptions({ editorProps: { attributes: ... } })` or `editor.view.setProps()`.

src/components/typist-editor.tsx
```diff
@@ -238,27 +230,32 @@ const TypistEditor = forwardRef<TypistEditorRef, TypistEditorProps>(function Typ
...
+                              placeholder: initialPlaceholder,
                           }),
                       ]
                     : []),
                 ExtraEditorCommands,
-                ViewEventHandlers.configure({
-                    onClick,
-                    onKeyDown,
-                }),
+                ViewEventHandlers,
```
doistbot: <big><kbd>🟡 P2</kbd></big>  `ViewEventHandlers` is now mounted without its initial `onClick`/`onKeyDown` options and only gets them later through `editor.storage.viewEventHandlers.setHandlers(...)` in an effect. That creates a first-commit gap where mount-time logic can't observe those handlers, and it also couples `TypistEditor` to the extension's private storage shape. Configure the extension with the initial handlers here, then keep the effect only for later prop updates.

src/components/typist-editor.tsx
```diff
@@ -321,62 +305,74 @@ const TypistEditor = forwardRef<TypistEditorRef, TypistEditorProps>(function Typ
...
-                },
-            },
-            extensions: allExtensions,
-            parseOptions: {
-                preserveWhitespace: isPlainTextDocument(schema),
+    const editor = useEditor({
+        autofocus: autoFocus ? 'end' : false,
+        content: htmlContent,
+        editable,
+        editorProps: {
```
doistbot: <big><kbd>🟡 P2</kbd></big>  These `editorProps`/`parseOptions` objects are rebuilt inline on every parent render. With Tiptap's built-in `useEditor`, changed option references cause the existing editor to be reconfigured in place, so unrelated parent rerenders now do extra ProseMirror work. Please memoize the nested option objects (or the full stable options shape) off `schema`/ARIA/editable so parent rerenders stay cheap.

src/components/typist-editor.test.tsx
```diff
@@ -216,4 +217,62 @@ describe('<TypistEditor />', () => {
             })
         })
     })
+
+    // The editor is created once and never recreated, so re-rendering with a new handler closure
+    // must still reach the latest handler rather than the one captured at mount.
+    describe('View Event Handlers', () => {
```
doistbot, line 223: <big><kbd>🟡 P2</kbd></big>  The new `View Event Handlers` block covers the handler-sync useEffect, but the other new runtime-sync path — `editable` applied via `editor.setEditable()` in a useEffect instead of by recreating the editor — is untested. The existing `aria-readonly` / read-only tests only pass `editable` at initial render, so a regression in the in-place toggle (e.g. the `syncEditableState` effect breaking or the dep being dropped) would slip through.

A focused test fits the harness already used here: render with `editable={true}`, `rerender` with `editable={false}`, and assert the editor actually toggles — either `typistEditorRef.current?.getEditor().isEditable === false` or the `contenteditable` attribute on the textbox. This mirrors the `rerender` + prop-change pattern the handler tests already establish.

doistbot (commented):
This PR replaces Typist's custom `useEditor` hook with Tiptap's built-in one, making the editor a stable instance for the component's lifetime and clearing the old `null`-editor crashes.

<sub>[Share Feedback](https://github.com/Doist/doistbot-app/issues/new?template=review-feedback.yml&title=Review+feedback+for+Doist%2Ftypist&pr_url=https%3A%2F%2Fgithub.com%2FDoist%2Ftypist%2Fpull%2F1387&datadog_url=https%3A%2F%2Fapp.datadoghq.com%2Flogs%3Fquery%3Dservice%253A%2528doistbot-app-sandbox%2520OR%2520doistbot-app%2529%2520%2540meta.deliveryId%253Ae8941690-6bc3-11f1-84d8-39a946869426%26agg_m%3Dcount%26agg_m_source%3Dbase%26agg_t%3Dcount%26clustering_pattern_field_path%3Dmessage%26cols%3Dhost%252Cservice%26fromUser%3Dtrue%26messageDisplay%3Dinline%26refresh_mode%3Dsliding%26storage%3Dhot%26stream_sort%3Ddesc%26viz%3Dstream%26live%3Dtrue) • [Review Logs](https://app.datadoghq.com/logs?query=service%3A%28doistbot-app-sandbox%20OR%20doistbot-app%29%20%40meta.deliveryId%3Ae8941690-6bc3-11f1-84d8-39a946869426&agg_m=count&agg_m_source=base&agg_t=count&clustering_pattern_field_path=message&cols=host%2Cservice&fromUser=true&messageDisplay=inline&refresh_mode=sliding&storage=hot&stream_sort=desc&viz=stream&live=true)</sub>

src/components/typist-editor.tsx
rfgamaral: Done. `editorProps` and `parseOptions` are now memoized (`editorProps` keyed on `schema`/`editable`/the aria props, `parseOptions` on `schema`), so they keep a stable reference across renders and the built-in `useEditor` no longer reconfigures ProseMirror on unrelated parent re-renders.

src/components/typist-editor.tsx
rfgamaral: Addressed together with the `editorProps` memoization. Worth noting `aria-readonly` was not actually going stale: with the inline object the built-in `useEditor` re-applied `editorProps` via `setOptions` on every render (the object always differed by reference), so the attribute was refreshed each time. The memoization just makes that efficient, and since `editable` is in the `editorProps` dependencies, toggling it still produces a fresh object and keeps `aria-readonly` in sync, so no separate sync effect is needed.

src/components/typist-editor.test.tsx
rfgamaral: Done. Added an `Editable State` test that renders with `editable={true}`, rerenders with `editable={false}`, and asserts both the `contenteditable` attribute and `getEditor().isEditable` flip. A regression in the `syncEditableState` effect (broken, or its `editable` dependency dropped) would fail it.

src/components/typist-editor.tsx
rfgamaral: Done. `ViewEventHandlers` is now configured with the initial `onClick`/`onKeyDown` (captured at mount), so the handlers are present from the first render and there's no mount-time gap. The effect is kept only to sync later handler changes.

On the storage coupling: I kept the runtime updates going through `setHandlers` deliberately. It's the only way to update the plugin's behavior on an editor that's created once while staying React-Compiler-safe and transaction-free, a command would dispatch an empty transaction, and passing a ref into `configure` is exactly the pattern the compiler rejects. The initial values now flow through the public `configure` options, and the component compiles cleanly under the React Compiler.

pedroalves0 (commented):
✅

PR #1394, opened 2026-06-24
feat(suggestion)!: read items on demand from a getter

rfgamaral:
## Overview

`createSuggestionExtension` previously captured its items when the extension was built and cached them in the editor storage. Since the editor is now created once and kept for the component's lifetime, an extension whose items come from a source that changes over time would be stuck with the items present when the editor first mounted.

This makes the factory's `items` parameter accept either a static array or a getter, and reads the current items on demand wherever the factory needs them: filtering the search, resolving the selected item, and rendering an already-inserted suggestion's label. A getter keeps suggestions in sync with a changing source without recreating the editor, mirroring how other runtime data is read reactively from inside an extension. The internal storage cache is removed, since the items are no longer held anywhere.

This is a breaking change: `onSearchChange` receives the current items array as its second argument instead of a `storage` object, and the `SuggestionStorage` type is no longer exported. Migrating is a direct swap from `storage.items` to `items`.

## Reference

- Builds on the editor lifecycle change in #1387

## PR Checklist

## Test plan

> [!TIP]
> Open the preview Storybook deployed to Netlify

- Type `@` followed by part of a name
    - [ ] Observe that matching suggestions appear and narrow as you keep typing
- Select a suggestion from the list
    - [ ] Observe that it is inserted with the correct name
- Type `#` followed by part of a hashtag
    - [ ] Observe that matching suggestions appear and can be selected

### DIFF
--- a/src/factories/create-suggestion-extension.ts
+++ b/src/factories/create-suggestion-extension.ts
@@ -93,7 +93,7 @@ type SuggestionOptions<TSuggestionItem> = {
      */
     onSearchChange?: (
         query: string,
-        storage: SuggestionStorage<TSuggestionItem>,
+        items: TSuggestionItem[],
     ) => TSuggestionItem[] | Promise<TSuggestionItem[]>
 
     /**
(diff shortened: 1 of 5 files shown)


nvignola (commented):
LGTM ✅

PR #1396, opened 2026-06-25
fix(typist-editor): correct editable state synchronization

rfgamaral:
## Overview

`TypistEditor` keeps the editor's `editable` state in sync through a React effect. This corrects two problems with that synchronization and tightens the typing around the internal view event handlers.

The effect previously ran after the browser painted, so switching the editor to read-only left a brief window where it was still writable and could process a queued keystroke or paste. Running the sync in a layout effect applies the change before paint and closes that window.

On mount the effect also re-applied the `editable` value the editor was already created with. That redundant call emitted an editor update before initialization finished, which arrived ahead of the editor's create event and could break extensions that set themselves up on creation. A guard now skips the call when the editable state already matches, so nothing runs on mount.

The internal access to the view event handlers storage is now typed instead of reaching through an untyped storage bag, so the contract is checked at compile time.

## PR Checklist

## Test plan

Code review and CI checks should be sufficient.

### DIFF
--- a/src/components/typist-editor.tsx
+++ b/src/components/typist-editor.tsx
@@ -1,4 +1,12 @@
-import { forwardRef, useCallback, useEffect, useImperativeHandle, useMemo, useState } from 'react'
+import {
+    forwardRef,
+    useCallback,
+    useEffect,
+    useImperativeHandle,
+    useLayoutEffect,
+    useMemo,
+    useState,
+} from 'react'
 
 import { getSchema } from '@tiptap/core'
 import { Placeholder } from '@tiptap/extension-placeholder'
@@ -390,7 +398,8 @@ const TypistEditor = forwardRef<TypistEditorRef, TypistEditorProps>(function Typ
         ...(onDestroy ? { onDestroy } : {}),
     })
 
-    useEffect(
+    // Apply editability before paint so a read-only editor can't process a queued edit first
+    useLayoutEffect(
         function syncEditableState() {
             editor.setEditable(editable)
         },
--- a/src/components/typist-editor.tsx
+++ b/src/components/typist-editor.tsx
@@ -398,10 +398,16 @@ const TypistEditor = forwardRef<TypistEditorRef, TypistEditorProps>(function Typ
         ...(onDestroy ? { onDestroy } : {}),
     })
 
-    // Apply editability before paint so a read-only editor can't process a queued edit first
+    // Sync editability in a layout effect so it applies before the browser paints, closing the gap
+    // where a read-only editor could still process a queued keystroke or paste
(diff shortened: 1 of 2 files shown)


doistbot (commented):
This PR fixes editable-state synchronization in `TypistEditor` by moving the sync to a layout effect, guarding against a redundant `setEditable` call on mount, and tightening typing for internal view-event-handler storage access.

Few things worth tightening:
- The `editor.storage.viewEventHandlers` cast still leaves the storage key lookup untyped, so renaming or removing the storage key won't be caught at compile time. Consider typing the storage on the editor side via Tiptap module augmentation or hiding it behind a small local helper rather than exporting the shape just to cast the bag here.

<details>

- **<a href="#">P3</a> src/components/typist-editor.tsx:418**: This cast still leaves the `editor.storage.viewEventHandlers` lookup itself untyped, so renaming the storage key or removing the extension won't be caught by the compiler. Since the codebase already uses Tiptap module augmentation for custom editor APIs, prefer typing the storage on the editor side (or hiding this behind a small local helper) instead of exporting the storage shape just to cast the bag here.

</details>

<sub>[Share Feedback](https://github.com/Doist/doistbot-app/issues/new?template=review-feedback.yml&title=Review+feedback+for+Doist%2Ftypist&pr_url=https%3A%2F%2Fgithub.com%2FDoist%2Ftypist%2Fpull%2F1396&datadog_url=https%3A%2F%2Fapp.datadoghq.com%2Flogs%3Fquery%3Dservice%253A%2528doistbot-app-sandbox%2520OR%2520doistbot-app%2529%2520%2540meta.deliveryId%253Aced7c670-7078-11f1-9e30-976ab68bbe19%26agg_m%3Dcount%26agg_m_source%3Dbase%26agg_t%3Dcount%26clustering_pattern_field_path%3Dmessage%26cols%3Dhost%252Cservice%26fromUser%3Dtrue%26messageDisplay%3Dinline%26refresh_mode%3Dsliding%26storage%3Dhot%26stream_sort%3Ddesc%26viz%3Dstream%26live%3Dtrue) • [Review Logs](https://app.datadoghq.com/logs?query=service%3A%28doistbot-app-sandbox%20OR%20doistbot-app%29%20%40meta.deliveryId%3Aced7c670-7078-11f1-9e30-976ab68bbe19&agg_m=count&agg_m_source=base&agg_t=count&clustering_pattern_field_path=message&cols=host%2Cservice&fromUser=true&messageDisplay=inline&refresh_mode=sliding&storage=hot&stream_sort=desc&viz=stream&live=true)</sub>

scottlovegrove (commented):
👍🏻

PR #1401, opened 2026-06-29
feat(typist-editor)!: adopt Tiptap's built-in `useEditor` hook

### LINKED ISSUE #532: Error:  Cannot read properties of null (reading 'updateOuterDeco')

### LINKED ISSUE #849: Switch custom useEditor hook for the official one provided by Tiptap

rfgamaral:
## Overview

Typist shipped a forked copy of Tiptap's `useEditor` hook that recreated the editor whenever props such as `content` or `extensions` changed. Recreating the editor mid-life was fragile and a recurring source of bugs. This change removes the fork and adopts Tiptap's built-in `useEditor`, so the editor is created once and kept for the lifetime of the component.

Because the editor is no longer recreated, `content`, `extensions`, and `placeholder` are now initialization-only: they're read when the editor first mounts and later changes to them are ignored. To change content at runtime, call `setContent()` on the editor instance; an extension that depends on changing data should stay stable and read that data on demand through a ref or store. The `editable` prop still updates at runtime, and event handler callbacks like `onUpdate` and `onFocus` always resolve to their latest version, so they no longer need to be wrapped in `useCallback`.

Suggestions follow the same model. `createSuggestionExtension` now accepts its items as either a static array or a getter that's read on demand, so suggestions stay in sync with a source that changes over time even though the editor is created only once. As part of this, `onSearchChange` receives the current items array as its second argument instead of a storage object, and the `SuggestionStorage` type is no longer exported.

New Storybook documentation covers the component lifecycle and how to keep UI outside the editor, such as a formatting toolbar, in sync by subscribing with `useSyncExternalStore` instead of re-rendering on every transaction.

## Reference

- Closes #849
- Fixes #532 and #737
- Promotes #1387, #1394, and #1396

## PR Checklist

## Test plan

These changes have been running in production across multiple public projects throughout the prerelease period, and each prerelease PR was tested on its own, so code review and CI checks should be sufficient here.


src/components/typist-editor.tsx
```diff
@@ -321,62 +322,98 @@ const TypistEditor = forwardRef<TypistEditorRef, TypistEditorProps>(function Typ
...
+                editor.setEditable(editable)
+            }
+        },
+        [editor, editable],
+    )
+
+    // The editor is created once, so push the latest handlers into the extension as they change
+    useEffect(
+        function syncViewEventHandlers() {
+            editor.commands.setViewEventHandlers({ onClick, onKeyDown })
```
doistbot, line 415: <a href="#">P2</a>  Using a command to update plugin storage triggers an empty ProseMirror transaction on every update. Since `onClick` and `onKeyDown` are often passed as inline functions, this can dispatch transactions and fire `onTransaction` listeners on every parent re-render. Mutate the storage directly instead:

```tsx
editor.storage.viewEventHandlers.onClick = onClick
editor.storage.viewEventHandlers.onKeyDown = onKeyDown
```

You can then remove the `setViewEventHandlers` command entirely.

src/factories/create-suggestion-extension.ts
```diff
@@ -225,14 +212,14 @@ function createSuggestionExtension<
                     default: null,
                     parseHTML: (element: Element) => {
                         const id = String(element.getAttribute('data-id'))
-                        const item = this.storage.itemsById[id]
+                        const item = findItemById(id)
```
doistbot, line 215: <a href="#">P1</a>  This lookup only runs while parsing HTML, not when an already-mounted suggestion node renders. With the editor now kept alive, if the items getter starts returning a renamed item (for example, a mention whose display name changed), the node still renders `node.attrs.label` from insertion time and stays stale until the document is reparsed. To make the new getter contract hold, existing suggestion nodes need a live update path (for example, updating their attrs from current items) instead of relying on `parseHTML`.

doistbot (commented):
This PR swaps Typist's forked `useEditor` for Tiptap's built-in version, giving the editor a stable single-instance lifecycle instead of recreating it on prop changes.

Few things worth tightening:

- **Event handler leak (P1):** Using conditional spread (`...(prop ? { prop } : {})`) for event handlers means Tiptap's `setOptions` ignores keys that are absent from the object, so old listeners never get unbound when a parent stops providing a handler. Pass each key explicitly so it's always present.
- **Suggestion node staleness (P1):** The items getter only runs during `parseHTML`, so already-mounted suggestion nodes keep their insertion-time `label` and don't update when item names change. Existing nodes need a live update path (e.g., updating attrs from current items) to honor the new dynamic-getter contract.
- **Stale handler window (P2):** `onClick`/`onKeyDown` are synced in a passive effect, leaving one frame where a swapped handler can still fire the previous closure. Moving to `useLayoutEffect` closes that gap.
- **Unnecessary transactions & N+1 lookups (P2):** The view-event-handler command dispatches an empty ProseMirror transaction on every parent re-render — mutate storage directly instead. Separately, `findItemById` does an `O(M)` scan per parsed/rendered node, causing `O(N*M)` complexity and calling the consumer's getter `N` times; caching the index by array reference keeps lookups `O(1)`. Also, `onItemSelect` re-resolves from `getItems()` at selection time, which can return a wrong item if the source changes between dropdown render and user pick — carry the selected item through the suggestion command instead.

<sub>[Share Feedback](https://github.com/Doist/doistbot-app/issues/new?template=review-feedback.yml&title=Review+feedback+for+Doist%2Ftypist&pr_url=https%3A%2F%2Fgithub.com%2FDoist%2Ftypist%2Fpull%2F1401&datadog_url=https%3A%2F%2Fapp.datadoghq.com%2Flogs%3Fquery%3Dservice%253A%2528doistbot-app-sandbox%2520OR%2520doistbot-app%2529%2520%2540meta.deliveryId%253Abf8c0720-73c6-11f1-9796-781a45c8f3d4%26agg_m%3Dcount%26agg_m_source%3Dbase%26agg_t%3Dcount%26clustering_pattern_field_path%3Dmessage%26cols%3Dhost%252Cservice%26fromUser%3Dtrue%26messageDisplay%3Dinline%26refresh_mode%3Dsliding%26storage%3Dhot%26stream_sort%3Ddesc%26viz%3Dstream%26live%3Dtrue) • [Review Logs](https://app.datadoghq.com/logs?query=service%3A%28doistbot-app-sandbox%20OR%20doistbot-app%29%20%40meta.deliveryId%3Abf8c0720-73c6-11f1-9796-781a45c8f3d4&agg_m=count&agg_m_source=base&agg_t=count&clustering_pattern_field_path=message&cols=host%2Cservice&fromUser=true&messageDisplay=inline&refresh_mode=sliding&storage=hot&stream_sort=desc&viz=stream&live=true)</sub>

src/components/typist-editor.tsx
```diff
@@ -321,62 +322,98 @@ const TypistEditor = forwardRef<TypistEditorRef, TypistEditorProps>(function Typ
...
+            // initializing (a tick later), breaking extensions that initialize with it.
+            if (editor.isEditable !== editable) {
+                editor.setEditable(editable)
+            }
+        },
+        [editor, editable],
+    )
+
+    // The editor is created once, so push the latest handlers into the extension as they change
+    useEffect(
```
doistbot: <a href="#">P2</a>  Updating `onClick`/`onKeyDown` in a passive effect leaves a post-commit window where the editor still calls the previous handlers. If a parent rerender swaps these props and the user clicks or types before this effect runs, one stale closure can fire even though this PR now promises latest-handler semantics. Syncing the storage in `useLayoutEffect` would close that gap.

src/components/typist-editor.tsx
```diff
@@ -321,62 +322,98 @@ const TypistEditor = forwardRef<TypistEditorRef, TypistEditorProps>(function Typ
...
+        // non-null editor instance from the first render.
+        immediatelyRender: true,
+
+        // Opt-out of the legacy behavior that re-renders the component on every ProseMirror
+        // transaction (e.g. keystrokes, selection changes). Typist doesn't read reactive editor
+        // state during render, so these re-renders would be wasteful. Consumers that need
+        // reactive state (e.g. toolbars) should subscribe directly via `useSyncExternalStore`.
+        shouldRerenderOnTransaction: false,
+
+        ...(onBeforeCreate ? { onBeforeCreate } : {}),
```
doistbot, line 375: <a href="#">P1</a>  Using conditional spread (`...(prop ? { prop } : {})`) for event handlers causes a listener leak. If a parent component initially provides an event handler (e.g., `onUpdate`) and later passes `undefined`, the property is omitted from the options object. Tiptap's `setOptions` iterates via `Object.keys(options)`, so it will ignore the missing key and never unbind the old listener, causing stale closures to execute. Pass the keys explicitly so they exist in the object (e.g., `onUpdate: onUpdate ? (props) => { ... } : undefined, onSelectionUpdate, onTransaction,`).

src/factories/create-suggestion-extension.ts
```diff
@@ -322,7 +305,7 @@ function createSuggestionExtension<
                             ])
                             .run()
 
-                        const item = storage.itemsById[props.id]
+                        const item = findItemById(String(props.id))
```
doistbot: <a href="#">P2</a>  `onItemSelect` is now re-resolved from `getItems()` at selection time. With the new getter/async model, the source can change after the dropdown rendered but before the user picks an item, so this callback can receive the wrong item or not fire at all. Carry the selected item (or a stable snapshot) through the suggestion command instead of doing a second lookup against mutable state.

src/factories/create-suggestion-extension.ts
```diff
@@ -178,8 +163,18 @@ function createSuggestionExtension<
     const idAttribute = String(attributesMapping[0]?.id ?? 'id')
     const labelAttribute = String(attributesMapping[0]?.label ?? 'label')
 
+    // Normalize the items into a getter, so a static array and a getter source are handled the
+    // same way internally. The getter is the single source of truth for the current items.
+    const getItems = typeof items === 'function' ? items : () => items
+
+    // Resolve a single item by its id from the current items. Used to look up the selected item
+    // and to resolve an inserted suggestion's label on demand, without keeping a stored index.
+    function findItemById(id: string): TSuggestionItem | undefined {
```
doistbot, line 172: <a href="#">P2</a>  N+1 performance trap: `findItemById` invokes `getItems()` and performs an `O(M)` array search. Because this is called for every parsed or rendered mention node during `getHTML()` and `parseHTML()`, it causes `O(N*M)` complexity and invokes the consumer's getter `N` times. If the consumer provides an expensive getter, this will block the main thread. Consider caching the index by the array reference to keep lookups `O(1)` while still supporting dynamic data updates:

```typescript
    let cachedItems: TSuggestionItem[] | undefined
    let cachedIndex: Map<string, TSuggestionItem> | undefined

    function findItemById(id: string): TSuggestionItem | undefined {
        const currentItems = getItems()
        if (currentItems !== cachedItems) {
            cachedItems = currentItems
            cachedIndex = new Map(currentItems.map(item => [String(item[idAttribute]), item]))
        }
        return cachedIndex!.get(id)
    }
```

rfgamaral: Keeping the plain `find` here. A few things make the index not worth it:

- `findItemById` only runs during `parseHTML` (and once on selection), not during `getHTML()`/rendering, which use the stored `node.attrs.label`. So there is no per-serialization N+1.
- An index would not reduce getter calls anyway: `getItems()` is invoked once per lookup regardless, so the expensive-getter concern is not addressed by indexing.
- At realistic sizes (a handful of nodes per parsed document over tens of items) the scan is negligible, while a reference-keyed index adds a staleness risk on in-place updates and does nothing for a getter that returns a fresh array each call.

The plain live `find` stays, it is always correct and cheap enough here.

src/components/typist-editor.tsx
rfgamaral: This does not leak in `@tiptap/react` 2.27.2. The handler options are not bound to the listener directly: `useEditor` constructs the editor once with stable wrapper functions, e.g. `onUpdate: (...args) => this.options.current.onUpdate?.(...args)`, and binds those once. Handler freshness does not come from `setOptions` rebinding (it only merges options), it comes from `this.options.current`, a ref that is fully replaced with the new options object on every render (`mostRecentOptions.current = options`).

So when a parent stops passing `onUpdate`, the omitted key makes `this.options.current.onUpdate` `undefined`, and the stable wrapper invokes it through optional chaining, which is a clean no-op. No stale closure fires, and nothing needs unbinding.

That also makes `...(onUpdate ? { onUpdate } : {})` and `onUpdate: onUpdate ? wrapper : undefined` functionally identical here, both leave `this.options.current.onUpdate` undefined and both no-op via the wrapper. Keeping the conditional spread.

src/factories/create-suggestion-extension.ts
rfgamaral: Addressed by resolving the picked item from the list the dropdown last rendered, rather than re-reading the live source, so the selected item always matches what the user clicked. Carrying the item through `command` would change the dropdown contract, which we want to avoid, so the snapshot stays internal. Async searches are guarded with a request id so a slow result cannot overwrite a newer one.

src/factories/create-suggestion-extension.ts
rfgamaral: This is by design, and it is not a regression from the getter contract. Suggestion nodes render from `node.attrs.label`, the value captured at insertion, and the label is only re-resolved from the source in the `label` attribute's `parseHTML`, which runs when content is parsed (load, `setContent`, paste). Mounted nodes have never updated live: before #1394 the same `parseHTML` resolved from the `itemsById` storage cache rather than a getter, and `renderHTML`/`renderText` read `node.attrs.label` then too. So the cache never drove live updates either.

What #1387 changed is only frequency: the editor used to be recreated on prop changes, which reparsed content and refreshed labels as a side effect; keeping it alive just makes that reparse rarer. The intended model is a snapshot at insertion that resolves from the source on the next parse and falls back to the stored label, which the getter does (and reads fresher data than the old cache). Live binding of mounted nodes would need a node view plus a re-render signal, a feature we do not need here.

I tightened the `parseHTML` comment so it no longer implies the label is always live.

rmartins90 (commented):
✅

PR #1403, opened 2026-06-30
docs: improve extensions guidance in component lifecycle

rfgamaral:
## Overview

Improves the Component Lifecycle docs on the `extensions` prop, with clearer guidance and examples for reading data that changes after mount.

## PR Checklist

## Test plan

Code review and CI checks should be sufficient.

### DIFF
--- a/stories/documentation/tips-and-tricks/component-lifecycle.md
+++ b/stories/documentation/tips-and-tricks/component-lifecycle.md
@@ -83,22 +83,21 @@ const extensions = useMemo(function configureExtensions() {
 }, [])
 ```
 
-Both the store getter and the seeded ref read fine while the editor is being created (in `onBeforeCreate` or `onCreate`): the store is readable at any time, and the ref already holds its seeded initial value.
+Both the store getter and the seeded ref can be read while the editor is being created, since the store is readable at any time and the ref already holds its seeded initial value.
 
 To skip maintaining the ref and effect yourself, you can reach for a `useEvent`-style hook (such as the [`react-use-event-hook`](https://npmx.dev/package/react-use-event-hook) package). It folds the same ref-and-effect mechanism into one stable function whose body always sees the latest props and state, and you pass it straight in as the getter.
 
-That convenience has one limit, though. A `useEvent`-style hook fills its internal ref from an effect, so unlike the store getter or the seeded ref it throws if called before mount. Typist sets `immediatelyRender: true`, so the editor is constructed during render, before that effect runs. If an extension reads the getter during editor construction, pass the initial value as a plain config option to cover that read and use the getter afterward. Reads after mount are safe: for example, `onCreate` is emitted from a `setTimeout` after the effect has filled the ref. Otherwise the getter alone is enough:
+It comes with one limit, though. The hook fills its internal ref from an effect, so it can't be read while the editor is being created. Typist sets `immediatelyRender: true`, so the editor is constructed during render, before that effect runs. Reach for a `useEvent`-style getter only when the extension reads the value after mount, and stick with the store or ref getter when it reads during construction:
 
 ```tsx
-// `initialCollaborators` covers the `onBeforeCreate` read; `getCollaborators` takes over afterwards.
-const [initialCollaborators] = useState(() => collaborators)
+// Do: a `useEvent` getter works for an extension that reads the value only after mount.
 const getCollaborators = useEvent(() => collaborators)
 
 const extensions = useMemo(
     function configureExtensions() {
-        return [MentionExtension.configure({ initialCollaborators, getCollaborators })]
+        return [MentionExtension.configure({ getCollaborators })]
     },
-    [initialCollaborators, getCollaborators],
+    [getCollaborators],
 )
 ```
 
--- a/stories/documentation/tips-and-tricks/component-lifecycle.md
+++ b/stories/documentation/tips-and-tricks/component-lifecycle.md
@@ -91,18 +91,14 @@ That convenience has one limit, though. A `useEvent`-style hook fills its intern
 
 ```tsx
 // `initialCollaborators` covers the `onBeforeCreate` read; `getCollaborators` takes over afterwards.
+const [initialCollaborators] = useState(() => collaborators)
 const getCollaborators = useEvent(() => collaborators)
 
 const extensions = useMemo(
     function configureExtensions() {
-        return [
-            MentionExtension.configure({
-                initialCollaborators: collaborators,
-                getCollaborators,
-            }),
-        ]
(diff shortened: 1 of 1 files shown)


stories/documentation/tips-and-tricks/component-lifecycle.md
```diff
@@ -49,17 +49,63 @@ const extensions = useMemo(
...
+const extensions = useMemo(
+    function configureExtensions() {
+        return [
+            MentionExtension.configure({
+                initialCollaborators: collaborators,
+                getCollaborators,
+            }),
+        ]
+    },
+    [getCollaborators],
```
doistbot: <a href="#">P2</a>  Omitting `collaborators` from the `useMemo` dependency array will trigger an `eslint-plugin-react-hooks/exhaustive-deps` warning for consumers copying this snippet. If they auto-fix the warning by adding it to the array, the `useMemo` will re-run on every change, defeating the purpose of keeping the extension instance stable.

Consider capturing the initial value with state so it can be safely included in the dependency array without causing re-evaluations:
```tsx
const [initialCollaborators] = useState(() => collaborators)
// ... pass initialCollaborators and include it in the deps array
```
Alternatively, an `// eslint-disable-next-line` comment could clarify that the omission is intentional.

stories/documentation/tips-and-tricks/component-lifecycle.md
```diff
@@ -49,17 +49,63 @@ const extensions = useMemo(
...
+    },
+    [collaborators],
+)
 
 const extensions = useMemo(function configureExtensions() {
     return [MentionExtension.configure({ getCollaborators: () => collaboratorsRef.current })]
 }, [])
 ```
 
+Both the store getter and the seeded ref read fine while the editor is being created (in `onBeforeCreate` or `onCreate`): the store is readable at any time, and the ref already holds its seeded initial value.
```
doistbot: <a href="#">P2</a>  The store-backed getter isn't safe for every creation-time read. `TypistEditor` builds the editor during render (`immediatelyRender: true`), so an extension that calls `getCollaborators()` from `onBeforeCreate` or another constructor-time path will read `store.getState()` during render. That bypasses React's subscription model and can initialize the editor from a stale store snapshot under concurrent rendering. Please limit this pattern to post-mount reads, or show the same `initialCollaborators + getter` split here when an extension needs data during construction.

doistbot (commented):
This PR improves the Component Lifecycle docs with clearer guidance on passing dynamic data to stable extension instances.

Few things worth tightening:
- The `useEvent` example omits `collaborators` from the `useMemo` dependency array, which will trigger an `exhaustive-deps` lint warning for consumers. If auto-fixed by adding it, the memo re-runs on every change and defeats the stability goal. Consider capturing the initial value with `useState` so it can safely live in the deps array, or add an eslint-disable comment to signal intent.
- The store-backed getter pattern isn't safe for construction-time reads. Since `TypistEditor` renders the editor immediately (`immediatelyRender: true`), an extension calling `getCollaborators()` from `onBeforeCreate` or similar constructor paths reads `store.getState()` during render, bypassing React's subscription model and risking stale snapshots under concurrent rendering. Worth noting this should be limited to post-mount reads, or showing the `initialCollaborators + getter` split for that case.
- The `useEvent` example recommends both `initialCollaborators` and `getCollaborators`, duplicating one piece of data across two config keys and forcing two code paths in the extension. Since the store/ref patterns above already handle pre-mount reads with a single getter API, consider framing `useEvent` as an option only when the getter is never read during construction rather than adding an extra `initial*` parameter.
- The npm link to `https://npmx.dev/package/react-use-event-hook` looks like a typo for `npmjs.com`.

<details>
<summary>Optional follow-up notes (2)</summary>

- **<a href="#">P3</a> stories/documentation/tips-and-tricks/component-lifecycle.md:88**: The URL `https://npmx.dev/package/react-use-event-hook` appears to have a typo in the domain (`npmx.dev` instead of `npmjs.com`).
- **<a href="#">P3</a> stories/documentation/tips-and-tricks/component-lifecycle.md:100**: This example asks extension authors to support both `initialCollaborators` and `getCollaborators`. That duplicates the same input in two config keys and forces the extension to maintain two code paths for one piece of data. Since the seeded-ref/store patterns above already handle pre-mount reads with a single getter API, I'd keep `useEvent` framed as an option only when the getter is never read during construction, rather than recommending an extra `initial*` parameter.

</details>

<sub>[Share Feedback](https://github.com/Doist/doistbot-app/issues/new?template=review-feedback.yml&title=Review+feedback+for+Doist%2Ftypist&pr_url=https%3A%2F%2Fgithub.com%2FDoist%2Ftypist%2Fpull%2F1403&datadog_url=https%3A%2F%2Fapp.datadoghq.com%2Flogs%3Fquery%3Dservice%253A%2528doistbot-app-sandbox%2520OR%2520doistbot-app%2529%2520%2540meta.deliveryId%253A3bd4b4a0-7474-11f1-8e31-dae10dbedfa9%26agg_m%3Dcount%26agg_m_source%3Dbase%26agg_t%3Dcount%26clustering_pattern_field_path%3Dmessage%26cols%3Dhost%252Cservice%26fromUser%3Dtrue%26messageDisplay%3Dinline%26refresh_mode%3Dsliding%26storage%3Dhot%26stream_sort%3Ddesc%26viz%3Dstream%26live%3Dtrue) • [Review Logs](https://app.datadoghq.com/logs?query=service%3A%28doistbot-app-sandbox%20OR%20doistbot-app%29%20%40meta.deliveryId%3A3bd4b4a0-7474-11f1-8e31-dae10dbedfa9&agg_m=count&agg_m_source=base&agg_t=count&clustering_pattern_field_path=message&cols=host%2Cservice&fromUser=true&messageDisplay=inline&refresh_mode=sliding&storage=hot&stream_sort=desc&viz=stream&live=true)</sub>

stories/documentation/tips-and-tricks/component-lifecycle.md
rfgamaral: The getter here is meant for on-demand reads (e.g. resolving suggestions when the user types `@`), which happen after mount and re-read the store each time, so a construction-time read doesn't get captured and go stale. The same render-time read also applies to the seeded ref (`collaboratorsRef.current`) shown just above as equally safe, so singling out the store would be inconsistent. The point this line is making is just the contrast that matters in practice: the store getter and the ref return a usable value during construction, while a `useEvent`-style getter throws. I'd rather keep that focused than fold in concurrent-rendering tearing caveats here. If an extension genuinely needs to capture a value at construction, the `initialCollaborators` split shown for the `useEvent` case covers that.

frankieyan (commented):
✅ Makes sense

PR #1415, opened 2026-07-14
feat(rich-text): preserve ordered list markers when unindenting out of a list

henningmu:
## Overview

Typing a number followed by a dot and a space at the start of a paragraph (e.g. the German date `17. Juli`) triggers the ordered-list input rule, which swallows the typed number into the list's `start` attribute. `Shift-Tab` to undo the list conversion, removed the number entirely (`17. Juli` → `Juli`), with no way to recover it. This made it effectively impossible to start a line with such a date.

This PR adds a `RichTextListItem` extension (mirroring the `RichTextOrderedList`/`RichTextBulletList` pattern) that replaces the stock `ListItem` in `RichTextKit`. It only customizes `Shift-Tab`: when the lift takes the item **out of an ordered list** (as opposed to unindenting it into a parent list), the rendered marker is re-inserted as literal text at the start of the lifted paragraph — so `Shift-Tab` never destroys visible content.

| Before | After |
| ------ | ----- |
| `17. Juli` + `Shift-Tab` → `Juli` | `17. Juli` + `Shift-Tab` → `17. Juli` |

Implementation notes:

Bullet lists are intentionally left untouched: the number is user-meaningful data, a bullet is not.

Related: Doist/Issues#20622

## PR Checklist

## Test plan

- Manually, in the preview Storybook:
  - Type `17. Juli` at the start of an empty line (converts to an ordered list starting at 17)
  - Press `Shift-Tab`
    - [ ] Observe the paragraph reads `17. Juli` as plain text
  - Press `Tab`/`Shift-Tab` inside nested and bullet lists
    - [ ] Observe unchanged behavior everywhere else

### DIFF
--- a/src/extensions/rich-text/rich-text-list-item.ts
+++ b/src/extensions/rich-text/rich-text-list-item.ts
@@ -1,0 +1,106 @@
+import { ListItem } from '@tiptap/extension-list-item'
+
+import type { Editor } from '@tiptap/core'
+
+/**
+ * Lifts the list item at the caret out of an ordered list while preserving the rendered list
+ * marker (e.g. `17.`) as literal text at the start of the lifted paragraph.
+ *
+ * This only applies when the lift would remove the item from the list altogether (i.e. the list
+ * is not nested within another list item), which is when the marker – rendered from the list's
+ * `start` attribute and the item's position – would otherwise be lost. The prime example is the
+ * automatic input rule for ordered lists swallowing a number the user meant as text (e.g. dates
+ * in German, like `17. Juli`), with `Shift-Tab` being the intuitive way to revert the conversion.
+ *
+ * @returns `true` if the marker-preserving lift was applied, `false` if the situation doesn't
+ * apply and the caller should fall back to the default behavior.
+ */
+function liftOutOfOrderedListPreservingMarker(
+    editor: Editor,
+    itemTypeName: string,
+    orderedListTypeName: string,
+): boolean {
+    const { selection } = editor.state
+
+    // Lifting a range of items merges them into a single item, making a single preserved marker
+    // ambiguous, so only handle a collapsed selection
+    if (!selection.empty) {
+        return false
+    }
+
+    const { $from } = selection
+
+    // Find the innermost list item the caret is placed within
+    let itemDepth: number | null = null
+    for (let depth = $from.depth; depth >= 2; depth--) {
+        if ($from.node(depth).type.name === itemTypeName) {
+            itemDepth = depth
+            break
+        }
+    }
+
+    if (itemDepth === null) {
+        return false
+    }
+
+    const list = $from.node(itemDepth - 1)
+
+    if (list.type.name !== orderedListTypeName) {
+        return false
+    }
+
+    // When the list is nested within another list item, lifting only unindents the item into the
+    // parent list (it remains a list item, and no marker is lost)
+    if ($from.node(itemDepth - 2).type.name === itemTypeName) {
+        return false
+    }
+
+    // Non-decimal markers (e.g. `<ol type="a">`) cannot be reconstructed as `<number>.` text
+    if (list.attrs.type) {
+        return false
+    }
+
+    if (!editor.can().liftListItem(itemTypeName)) {
+        return false
+    }
+
+    // The number the item is rendered with, given the list start and the item's position
+    const marker = `${Number(list.attrs.start ?? 1) + $from.index(itemDepth - 1)}. `
+
+    // The position at the start of the item's first paragraph, where the marker text belongs
+    const markerPosition = $from.start(itemDepth) + 1
+
+    return editor
+        .chain()
+        .liftListItem(itemTypeName)
+        .command(({ tr }) => {
+            tr.insertText(marker, tr.mapping.map(markerPosition))
+            return true
+        })
+        .run()
+}
+
+/**
+ * Custom extension that extends the built-in `ListItem` extension to preserve the rendered list
+ * marker as literal text when `Shift-Tab` lifts an item out of an ordered list, making sure that
+ * a number the user typed themselves – swallowed into the list's `start` attribute by the
+ * automatic input rule – is never lost.
+ */
+const RichTextListItem = ListItem.extend({
+    addKeyboardShortcuts() {
+        return {
+            ...this.parent?.(),
+            'Shift-Tab': () => {
+                return (
+                    liftOutOfOrderedListPreservingMarker(
+                        this.editor,
+                        this.name,
+                        this.options.orderedListTypeName,
+                    ) || this.editor.commands.liftListItem(this.name)
+                )
+            },
+        }
+    },
+})
+
+export { RichTextListItem }
--- a/src/extensions/rich-text/rich-text-list-item.ts
+++ b/src/extensions/rich-text/rich-text-list-item.ts
@@ -1,3 +1,4 @@
+import { findParentNode } from '@tiptap/core'
 import { ListItem } from '@tiptap/extension-list-item'
 
 import type { Editor } from '@tiptap/core'
@@ -30,20 +31,13 @@ function liftOutOfOrderedListPreservingMarker(
 
     const { $from } = selection
 
-    // Find the innermost list item the caret is placed within
-    let itemDepth: number | null = null
-    for (let depth = $from.depth; depth >= 2; depth--) {
-        if ($from.node(depth).type.name === itemTypeName) {
-            itemDepth = depth
-            break
-        }
-    }
+    const item = findParentNode((node) => node.type.name === itemTypeName)(selection)
 
-    if (itemDepth === null) {
+    if (!item || item.depth < 2) {
         return false
     }
 
-    const list = $from.node(itemDepth - 1)
+    const list = $from.node(item.depth - 1)
 
     if (list.type.name !== orderedListTypeName) {
         return false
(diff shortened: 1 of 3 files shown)


src/extensions/rich-text/rich-text-list-item.ts
```diff
@@ -0,0 +1,106 @@
...
+function liftOutOfOrderedListPreservingMarker(
+    editor: Editor,
+    itemTypeName: string,
+    orderedListTypeName: string,
+): boolean {
+    const { selection } = editor.state
+
+    // Lifting a range of items merges them into a single item, making a single preserved marker
+    // ambiguous, so only handle a collapsed selection
+    if (!selection.empty) {
```
doistbot, line 28: <a href="#">P2</a>  This falls back to the stock lift for any non-empty selection, including a selection entirely within one list item where the marker is unambiguous. For example, selecting `Juli` in `17. Juli` and pressing Shift-Tab produces `Juli`, losing `17.` again. Preserve the marker when the selected range covers exactly one item; only skip the multi-item case.

src/extensions/rich-text/rich-text-list-item.ts
```diff
@@ -0,0 +1,106 @@
...
+        .run()
+}
+
+/**
+ * Custom extension that extends the built-in `ListItem` extension to preserve the rendered list
+ * marker as literal text when `Shift-Tab` lifts an item out of an ordered list, making sure that
+ * a number the user typed themselves – swallowed into the list's `start` attribute by the
+ * automatic input rule – is never lost.
+ */
+const RichTextListItem = ListItem.extend({
```
doistbot, line 83: <a href="#">P1</a>  The `this.options.orderedListTypeName` property is undefined at runtime and causes a TypeScript compilation error because it is not declared as a valid option on `ListItem` or `RichTextListItem`.

To resolve this, define a custom options interface and provide a default value in `addOptions()`:

```typescript
import type { ListItemOptions } from '@tiptap/extension-list-item'

export interface RichTextListItemOptions extends ListItemOptions {
    orderedListTypeName: string
}

const RichTextListItem = ListItem.extend<RichTextListItemOptions>({
    addOptions() {
        return {
            ...this.parent?.(),
            orderedListTypeName: 'orderedList',
        }
    },
    // ...
})
```

Additionally, update `listItem` in `RichTextKitOptions` (`src/extensions/rich-text/rich-text-kit.ts`) to use `Partial<RichTextListItemOptions> | false`.

src/extensions/rich-text/rich-text-list-item.ts
```diff
@@ -0,0 +1,106 @@
...
+    if (list.attrs.type) {
+        return false
+    }
+
+    if (!editor.can().liftListItem(itemTypeName)) {
+        return false
+    }
+
+    // The number the item is rendered with, given the list start and the item's position
+    const marker = `${Number(list.attrs.start ?? 1) + $from.index(itemDepth - 1)}. `
```
doistbot: <a href="#">P1</a>  The marker calculation uses `$from.index(itemDepth - 1)`, which returns the index of the **list node** within its parent (the document), not the index of the **list item** within the list. For a single-item list both are 0 by coincidence, but for any item beyond the first the marker will always be `start + 0` instead of `start + itemPosition`.

```typescript
// Current: index of the orderedList within the doc (always 0 for a top-level list)
const marker = `${Number(list.attrs.start ?? 1) + $from.index(itemDepth - 1)}. `

// Should be: index of the listItem within the orderedList
const marker = `${Number(list.attrs.start ?? 1) + $from.index(itemDepth)}. `
```

The multi-item test (`preserves the rendered marker matching the item position within the list`) expects `2. two` for the second item — if this test passes, double-check that it's actually exercising the code path (e.g. the assertion might be matching a different output).

doistbot (commented):
This PR adds a `RichTextListItem` extension that preserves ordered-list markers as literal text when `Shift-Tab` lifts an item out of a list — a nice fix for the date-at-start-of-line problem.

<details>
<summary>Optional follow-up notes (3)</summary>

- **<a href="#">P3</a> src/extensions/rich-text/rich-text-list-item.ts:63**: The `editor.can().liftListItem(itemTypeName)` guard at this line dry-runs the lift command's transaction logic to check feasibility, then `editor.chain().liftListItem(itemTypeName)` at line 75 runs the same logic again to actually apply it. The chain already handles failure — if `liftListItem` returns false, the chain returns false and the fallback `this.editor.commands.liftListItem(this.name)` on line 99 fires correctly. The guard adds no safety benefit (the lift happens before marker insertion, so there's no risk of partial execution) and doubles the lift computation on every Shift-Tab that reaches this path. Remove lines 63–65.
- **<a href="#">P3</a> src/extensions/rich-text/rich-text-list-item.test.ts:83**: The describe block name says "when `Shift-Tab` does not lift the item out of a list," but the bullet-list test inside it (line 84) *does* lift the item out — it goes from `<ul><li>` to plain `<p>`. The describe name is only accurate for the nested ordered list test. Consider renaming to something like `when marker preservation is not applicable` or `when `Shift-Tab` falls back to the default behavior`.
- **<a href="#">P3</a> src/extensions/rich-text/rich-text-list-item.test.ts:47**: The non-empty selection guard in `liftOutOfOrderedListPreservingMarker` (line 27 of `rich-text-list-item.ts`) is a deliberate safety behavior — it prevents ambiguous marker insertion when lifting a multi-item range — but no test covers it. Every test uses `createEditorWithCaretAfter`, which always creates a collapsed selection. A test that selects across two ordered-list items (e.g. via `editor.commands.setTextSelection(from, to)`) and asserts that Shift-Tab lifts without inserting a stray marker would protect against someone removing the guard. The existing harness makes this straightforward to add.

</details>

<sub>[Share Feedback](https://github.com/Doist/doistbot-app/issues/new?template=review-feedback.yml&title=Review+feedback+for+Doist%2Ftypist&pr_url=https%3A%2F%2Fgithub.com%2FDoist%2Ftypist%2Fpull%2F1415&datadog_url=https%3A%2F%2Fapp.datadoghq.com%2Flogs%3Fquery%3Dservice%253A%2528doistbot-app-sandbox%2520OR%2520doistbot-app%2529%2520%2540meta.deliveryId%253A5cc5f4c0-7f7e-11f1-8502-09264b36622f%26agg_m%3Dcount%26agg_m_source%3Dbase%26agg_t%3Dcount%26clustering_pattern_field_path%3Dmessage%26cols%3Dhost%252Cservice%26fromUser%3Dtrue%26messageDisplay%3Dinline%26refresh_mode%3Dsliding%26storage%3Dhot%26stream_sort%3Ddesc%26viz%3Dstream%26live%3Dtrue) • [Review Logs](https://app.datadoghq.com/logs?query=service%3A%28doistbot-app-sandbox%20OR%20doistbot-app%29%20%40meta.deliveryId%3A5cc5f4c0-7f7e-11f1-8502-09264b36622f&agg_m=count&agg_m_source=base&agg_t=count&clustering_pattern_field_path=message&cols=host%2Cservice&fromUser=true&messageDisplay=inline&refresh_mode=sliding&storage=hot&stream_sort=desc&viz=stream&live=true)</sub>

src/extensions/rich-text/rich-text-list-item.ts
henningmu: Wrong. `orderedListTypeName` is already defined and defaulted by Tiptap’s `ListItemOptions`; TypeScript passes.

src/extensions/rich-text/rich-text-list-item.ts
henningmu: Wrong. `$from.index(itemDepth - 1)` correctly returns the list item’s index. The existing second-item regression test passes; changing it to `itemDepth` would break marker numbering.

src/extensions/rich-text/rich-text-list-item.ts
```diff
@@ -0,0 +1,106 @@
...
+    const { $from } = selection
+
+    // Find the innermost list item the caret is placed within
+    let itemDepth: number | null = null
+    for (let depth = $from.depth; depth >= 2; depth--) {
+        if ($from.node(depth).type.name === itemTypeName) {
+            itemDepth = depth
+            break
+        }
+    }
```
rfgamaral: Could we use Tiptap’s `findParentNode` helper here? It performs the same closest-parent walk from `selection.$from` and returns both `depth` and `start`, removing the custom traversal:

```ts
const item = findParentNode((node) => node.type.name === itemTypeName)(selection)

if (!item || item.depth < 2) {
    return false
}

const list = $from.node(item.depth - 1)
```

The remaining references become `item.depth`, and `markerPosition` becomes `item.start + 1`. The depth guard preserves the loop’s current `depth >= 2` boundary. This seems easier to maintain.

src/extensions/rich-text/rich-text-list-item.ts
```diff
@@ -0,0 +1,106 @@
...
+
+    // When the list is nested within another list item, lifting only unindents the item into the
+    // parent list (it remains a list item, and no marker is lost)
+    if ($from.node(itemDepth - 2).type.name === itemTypeName) {
+        return false
+    }
+
+    // Non-decimal markers (e.g. `<ol type="a">`) cannot be reconstructed as `<number>.` text
+    if (list.attrs.type) {
```
rfgamaral: Small compatibility edge case: pasted or imported HTML can contain `<ol type="1">`, where `1` still represents decimal numbering. Tiptap preserves that attribute, but this check treats it as non-decimal, so `Shift-Tab` still drops the marker. Could we only fall back for genuinely non-decimal types?

```ts
if (list.attrs.type && list.attrs.type !== '1') {
    return false
}
```

rfgamaral (commented):
@henningmu Looks good to me, left a few comments that I think are worth addressing, but non-blocking.

src/extensions/rich-text/rich-text-list-item.ts
henningmu: Done 👍

src/extensions/rich-text/rich-text-list-item.ts
henningmu: Done 👍

henningmu:
Addressed both comments. Thanks, Ricardo 🙌

rfgamaral (approved):

PR #1441, opened 2026-07-27
feat: support React 19

rfgamaral:
## Overview

Widens the React peer range to `^18.0.0 || ^19.0.0` and moves the development environment to React 19.

Today `@doist/typist` peers on React 18 only, which blocks consumers from even trying React 19: npm fails to resolve. `@doist/reactist` already allows both (`>=18.0.0 <20.0.0`), so Typist is the remaining gate. Both `todoist-web` and `comms-web` are still on React 18.3.1, so nothing changes for them right now.

Typist barely touches React — a single `forwardRef` component and a handful of hooks, with Tiptap handling the editor integration, and `@tiptap/react` already supports 17, 18, and 19. The built output is byte-for-byte identical whether it is compiled against React 18 or 19 types.

Since development now happens on React 19, a `React 18 Compatibility` job installs React 18 over the top of a normal install and runs the same type checking and test suite against it. That keeps the oldest supported version covered without aliased packages, extra configs, or duplicated scripts. When React 18 support is eventually dropped, deleting that job is the only cleanup needed. The job restores the shared `node_modules` cache but never saves it, so the downgraded tree cannot leak into the other jobs.

Replaces [#1325](https://github.com/Doist/typist/pull/1325), which bumped the same dependencies but left React 18 untested, and would not have published the widened range from a `chore` commit.

## PR Checklist

## Test plan

- Observe that CI is green, which covers static analysis, tests, and the build on React 19, plus type checking and tests on React 18
- Open the Netlify preview Storybook and confirm the editor stories behave as expected under React 19

### DIFF
--- a/.github/workflows/check-ci-validation.yml
+++ b/.github/workflows/check-ci-validation.yml
@@ -122,6 +122,50 @@ jobs:
               run: |
                   npm run test
 
+    react-18-compatibility:
+        name: React 18 Compatibility
+        runs-on: ubuntu-latest
+        timeout-minutes: 15
+
+        needs:
+            - prepare-workflow
+
+        steps:
+            - name: Checkout repository
+              uses: actions/checkout@v6
+
+            - name: Prepare Node.js environment
+              uses: actions/setup-node@v6
+              with:
+                  cache: npm
+                  node-version-file: .node-version
+
+            # The cache is only restored, and never saved, because this job replaces
+            # React with an older version that the other jobs must not inherit.
+            - name: Restore project 'node_modules' directory
+              id: node-modules-cache
+              uses: actions/cache/restore@v5
+              with:
+                  key: node-modules-cache-${{ hashFiles('**/package-lock.json', '**/.node-version') }}
+                  path: node_modules/
+
+            - name: Install project npm dependencies
+              if: ${{ steps.node-modules-cache.outputs.cache-hit != 'true' }}
+              run: |
+                  npm ci
+
+            - name: Downgrade React to the oldest supported version
+              run: |
+                  npm install --no-save react@18 react-dom@18 @types/react@18 @types/react-dom@18
+
+            - name: Perform type checking with TypeScript
+              run: |
+                  npm run check:types
+
+            - name: Test codebase correctness
+              run: |
+                  npm run test
+
     build-package:
         name: Build Package
         runs-on: ubuntu-latest
--- a/.github/workflows/check-ci-validation.yml
+++ b/.github/workflows/check-ci-validation.yml
@@ -154,9 +154,11 @@ jobs:
               run: |
                   npm ci
 
+            # React 18 is frozen at its final release, while its types still receive
+            # patches, so only the runtime packages are pinned to an exact version
             - name: Downgrade React to the oldest supported version
               run: |
-                  npm install --no-save react@18 react-dom@18 @types/react@18 @types/react-dom@18
+                  npm install --no-save react@18.3.1 react-dom@18.3.1 @types/react@18 @types/react-dom@18
 
             - name: Perform type checking with TypeScript
               run: |
--- a/package-lock.json
+++ b/package-lock.json
@@ -77,8 +77,8 @@
                 "@testing-library/user-event": "14.6.1",
                 "@types/hast": "3.0.5",
                 "@types/lodash-es": "4.17.12",
-                "@types/react": "18.3.31",
-                "@types/react-dom": "18.3.7",
+                "@types/react": "19.2.17",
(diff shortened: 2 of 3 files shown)


.github/workflows/check-ci-validation.yml
```diff
@@ -122,6 +122,50 @@ jobs:
...
+                  path: node_modules/
+
+            - name: Install project npm dependencies
+              if: ${{ steps.node-modules-cache.outputs.cache-hit != 'true' }}
+              run: |
+                  npm ci
+
+            - name: Downgrade React to the oldest supported version
+              run: |
+                  npm install --no-save react@18 react-dom@18 @types/react@18 @types/react-dom@18
```
doistbot: <a href="#">P2</a>  `react@18` and the other `@18` specs resolve to the latest React 18 releases, not the oldest versions allowed by `^18.0.0`. This leaves consumers on React 18.0.x untested despite the new job's stated purpose. Pin the compatibility dependencies to the intended minimum versions (for example, `react@18.0.0` and `react-dom@18.0.0`), or rename the job/comment to describe latest-React-18 coverage.

doistbot (commented):
Widens the React peer range to `^18.0.0 || ^19.0.0`, moves the dev environment to React 19, and adds a well-constructed restore-only-cache CI job to keep React 18 type-checking and tests covered.

Few things worth tightening:
- Pin the compatibility job's React 18 packages to the actual minimum versions (e.g. `react@18.0.0`, `react-dom@18.0.0`) — bare `@18` specs resolve to the latest 18 release, so 18.0.x consumers go untested despite the job's stated purpose.
- After the single React 18 install, run type checking and tests concurrently with `run-p check:types test` to shorten the job's wall-clock time.

<details>

- **<a href="#">P3</a> .github/workflows/check-ci-validation.yml:161**: `check:types` has `noEmit` and the test script does not depend on it, so these two validations can run concurrently after the single React 18 install. Running them with `run-p check:types test` would shorten the compatibility job's wall-clock time without duplicating setup work.

</details>

<sub>[Share Feedback](https://github.com/Doist/doistbot-app/issues/new?template=review-feedback.yml&title=Review+feedback+for+Doist%2Ftypist&pr_url=https%3A%2F%2Fgithub.com%2FDoist%2Ftypist%2Fpull%2F1441&datadog_url=https%3A%2F%2Fapp.datadoghq.com%2Flogs%3Fquery%3Dservice%253A%2528doistbot-app-sandbox%2520OR%2520doistbot-app%2529%2520%2540meta.deliveryId%253Ae4a9dba0-89e1-11f1-8ead-08ba6e8fa200%26agg_m%3Dcount%26agg_m_source%3Dbase%26agg_t%3Dcount%26clustering_pattern_field_path%3Dmessage%26cols%3Dhost%252Cservice%26fromUser%3Dtrue%26messageDisplay%3Dinline%26refresh_mode%3Dsliding%26storage%3Dhot%26stream_sort%3Ddesc%26viz%3Dstream%26live%3Dtrue) • [Review Logs](https://app.datadoghq.com/logs?query=service%3A%28doistbot-app-sandbox%20OR%20doistbot-app%29%20%40meta.deliveryId%3Ae4a9dba0-89e1-11f1-8ead-08ba6e8fa200&agg_m=count&agg_m_source=base&agg_t=count&clustering_pattern_field_path=message&cols=host%2Cservice&fromUser=true&messageDisplay=inline&refresh_mode=sliding&storage=hot&stream_sort=desc&viz=stream&live=true)</sub>

.github/workflows/check-ci-validation.yml
rfgamaral: Pinned the runtime packages to `react@18.3.1` and `react-dom@18.3.1` in 49e24b9, so the job now installs an exact version instead of resolving the newest 18.x.

Kept the types on `@18`, since `@types/react` 18.x still receives patches and the latest of those is what a consumer on React 18 would actually install. Did not go down to `18.0.0`: the React 18 line ended at 18.3.1, `npm install react@18` gives 18.3.1, and both consumers run that exact version. 18.0.0 also predates `React.act`, so Testing Library would fall back to the deprecated `react-dom/test-utils` path and we would be chasing warnings from a release nobody uses.

rfgamaral:
On the optional note about running the two validations concurrently with `run-p check:types test`: skipping it. Type checking takes ~0.5s and the tests ~1.4s, against a job whose runtime is dominated by the install, so the gain is under a second. Not worth interleaving the output and making failures harder to read.

frankieyan (commented):
I like it! Cleaner than https://github.com/Doist/reactist/pull/1074 🎊

PR #1466, opened 2026-08-11
fix(serializers): preserve angle-bracket URLs as text

rfgamaral:
## Overview

Treats angle-bracket HTTP(S) URLs as plain text when loading stored Markdown into rich-text editors. This prevents unsupported Markdown syntax from becoming a link whether or not the link extension is enabled.

## Reference

Related to [Doist/Issues#20812](https://github.com/Doist/Issues/issues/20812)

## PR Checklist

## Test plan

1. Open the deployed Storybook preview
2. Open the `Default` story under `Typist Editor → Rich-text`
3. Set the `content` control to `<https://duckduckgo.com>`
    - [ ] Observe that the editor displays `<https://duckduckgo.com>` as plain text rather than a link
    - [ ] Observe that `Markdown Output` contains `<https://duckduckgo.com>` unchanged

## Demo

| image | image |

### DIFF
--- a/src/serializers/html/plugins/remark-ignore-angle-bracket-autolinks.ts
+++ b/src/serializers/html/plugins/remark-ignore-angle-bracket-autolinks.ts
@@ -1,0 +1,36 @@
+import { visit } from 'unist-util-visit'
+
+import type { Transformer } from 'unified'
+import type { Parent } from 'unist'
+
+/**
+ * Treats angle-bracket HTTP(S) autolinks as plain text because they are not a supported rich-text
+ * Markdown shortcut.
+ */
+function remarkIgnoreAngleBracketAutolinks(): Transformer {
+    return (tree, file) => {
+        visit(tree as Parent, (node, index, parent) => {
+            const startOffset = node.position?.start.offset
+            const endOffset = node.position?.end.offset
+
+            if (
+                node.type !== 'link' ||
+                typeof startOffset !== 'number' ||
+                typeof endOffset !== 'number' ||
+                typeof index !== 'number' ||
+                !parent
+            ) {
+                return
+            }
+
+            const source = String(file.value).slice(startOffset, endOffset)
+
+            if (/^<https?:\/\/\S+>$/i.test(source)) {
+                const textNode = { type: 'text', value: source }
+                parent.children[index] = textNode
+            }
+        })
+    }
+}
+
+export { remarkIgnoreAngleBracketAutolinks }
--- a/src/serializers/html/plugins/remark-ignore-angle-bracket-autolinks.ts
+++ b/src/serializers/html/plugins/remark-ignore-angle-bracket-autolinks.ts
@@ -1,15 +1,15 @@
 import { visit } from 'unist-util-visit'
 
+import type { Root } from 'mdast'
 import type { Transformer } from 'unified'
-import type { Parent } from 'unist'
 
 /**
  * Treats angle-bracket HTTP(S) autolinks as plain text because they are not a supported rich-text
  * Markdown shortcut.
  */
-function remarkIgnoreAngleBracketAutolinks(): Transformer {
+function remarkIgnoreAngleBracketAutolinks(): Transformer<Root> {
     return (tree, file) => {
-        visit(tree as Parent, (node, index, parent) => {
+        visit(tree, 'link', (node, index, parent) => {
             const startOffset = node.position?.start.offset
             const endOffset = node.position?.end.offset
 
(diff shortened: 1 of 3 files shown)


src/serializers/html/plugins/remark-ignore-angle-bracket-autolinks.ts
```diff
@@ -0,0 +1,36 @@
+import { visit } from 'unist-util-visit'
+
+import type { Transformer } from 'unified'
+import type { Parent } from 'unist'
+
+/**
+ * Treats angle-bracket HTTP(S) autolinks as plain text because they are not a supported rich-text
+ * Markdown shortcut.
+ */
+function remarkIgnoreAngleBracketAutolinks(): Transformer {
```
doistbot: <a href="#">P2</a>  The existing `remarkDisableConstructs` plugin already uses micromark's `disable` extension to turn off constructs at parse time (e.g. `labelStartLink`, `codeText`, `headingAtx`). Adding `'autolink'` to its `disabledConstructs` list would prevent angle-bracket URLs from being parsed as links in the first place — no AST walk, source-text slicing, or regex needed. The `autolink` construct also covers email autolinks (`<foo@bar.com>`), which is consistent with the stated goal of treating angle-bracket autolinks as an unsupported shortcut. Since the behavior should apply unconditionally (the PR targets both the with-link and without-link cases), the push would go outside the schema conditionals in that plugin.

doistbot (commented):
The PR treats angle-bracket HTTP(S) URLs as plain text when loading stored Markdown into rich-text editors, preventing them from being parsed as links.

Few things worth tightening:
- Consider reusing the existing `remarkDisableConstructs` plugin instead of the new AST-walking plugin. Adding `'autolink'` to its `disabledConstructs` list would disable angle-bracket URL parsing at the micromark level—no AST traversal, source slicing, or regex needed. Since the behavior applies unconditionally, the push should go outside the schema conditionals in that plugin.

I also left one optional follow-up note in the details below.

<details>

- **<a href="#">P3</a> src/serializers/html/plugins/remark-ignore-angle-bracket-autolinks.ts:12**: `visit(tree as Parent, ...)` invokes the callback for every node in the AST, then early-returns for non-`link` nodes after accessing position properties. Every other visitor in this directory uses the typed form (e.g. `visit(tree, 'element', ...)`) to skip non-matching nodes. Use `visit(tree, 'link', (node, index, parent) => { ... })` to avoid the per-node callback overhead on the serialization hot path, which runs on every document load and scales with document size.

</details>

<sub>[Share Feedback](https://github.com/Doist/doistbot-app/issues/new?template=review-feedback.yml&title=Review+feedback+for+Doist%2Ftypist&pr_url=https%3A%2F%2Fgithub.com%2FDoist%2Ftypist%2Fpull%2F1466&datadog_url=https%3A%2F%2Fapp.datadoghq.com%2Flogs%3Fquery%3Dservice%253A%2528doistbot-app-sandbox%2520OR%2520doistbot-app%2529%2520%2540meta.deliveryId%253A75572e30-95ab-11f1-9ce9-42281f1c8cc5%26agg_m%3Dcount%26agg_m_source%3Dbase%26agg_t%3Dcount%26clustering_pattern_field_path%3Dmessage%26cols%3Dhost%252Cservice%26fromUser%3Dtrue%26messageDisplay%3Dinline%26refresh_mode%3Dsliding%26storage%3Dhot%26stream_sort%3Ddesc%26viz%3Dstream%26live%3Dtrue) • [Review Logs](https://app.datadoghq.com/logs?query=service%3A%28doistbot-app-sandbox%20OR%20doistbot-app%29%20%40meta.deliveryId%3A75572e30-95ab-11f1-9ce9-42281f1c8cc5&agg_m=count&agg_m_source=base&agg_t=count&clustering_pattern_field_path=message&cols=host%2Cservice&fromUser=true&messageDisplay=inline&refresh_mode=sliding&storage=hot&stream_sort=desc&viz=stream&live=true)</sub>

src/serializers/html/plugins/remark-ignore-angle-bracket-autolinks.ts
rfgamaral: @doistbot Disabling the CommonMark `autolink` construct does not work while `remarkAutolinkLiteral` is enabled. The opening bracket becomes text and the GFM extension relinks the inner URL, producing `<p>&lt;<a href="https://duckduckgo.com%3E">https://duckduckgo.com></a></p>` for `<https://duckduckgo.com>`. We need the source-aware transform to prevent that relinking.

doistbot: You’re right—thanks for checking this. I was treating CommonMark’s `autolink` as the only parsing path, but with `remarkAutolinkLiteral` enabled, disabling it leaves `<` as text and lets GFM relink `https://duckduckgo.com>`, which explains the encoded `>` in the URL.

The source-aware transform is therefore the correct approach: it can identify the original angle-bracket syntax and replace the parsed link with its literal source text. I withdraw the earlier suggestion to add `autolink` to `remarkDisableConstructs`; the tests covering schemas both with and without the link mark should remain to protect these two parsing paths.

